PR #154, opened 2025-09-04
when one/more RPC nodes are down, still return successful api responses for the other networks

chimalsky:
Fixes notion task: https://www.notion.so/d05e7f2ecd1e404eb4dfdcd61ad01e8e?v=78fd36915e0f4f9e8ba7553b76aa815f&p=245ae642e01880ffb62eff7e4d7c304b&pm=c

This PR aims to keep code diff to minimum. It aims to convert the optimistic `Promise.all()` to use `Promise.allSettled()` in order to allow for moonwell-sdk client to return the successfully responses for network(s) requests when one or more of the configured network(s)' RPC node is down.

### DIFF
--- a/src/actions/core/user-positions/common.ts
+++ b/src/actions/core/user-positions/common.ts
@@ -11,101 +11,96 @@ export const getUserPositionData = async (params: {
 }) => {
   const viewsContract = params.environment.contracts.views;
 
-  const userData = await Promise.allSettled([
-    viewsContract?.read.getAllMarketsInfo(),
-    viewsContract?.read.getUserBalances([params.account]),
-    viewsContract?.read.getUserBorrowsBalances([params.account]),
-    viewsContract?.read.getUserMarketsMemberships([params.account]),
-  ]);
-
-  const [allMarketsRes, balancesRes, borrowsRes, membershipsRes] = userData;
-
-  const allMarkets =
-    allMarketsRes.status === "fulfilled" ? allMarketsRes.value : undefined;
-  const balances =
-    balancesRes.status === "fulfilled" ? balancesRes.value : undefined;
-  const borrows =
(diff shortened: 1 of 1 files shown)


src/actions/core/user-positions/common.ts
```diff
@@ -11,14 +11,27 @@ export const getUserPositionData = async (params: {
 }) => {
   const viewsContract = params.environment.contracts.views;
 
-  const userData = await Promise.all([
+  const userData = await Promise.allSettled([
```
x0s0l, line 14: IIRC if you use Promise.all on viem, it will use a batch call (multicall) so all those subsequent calls will either fail or resolve

No need to use allSettled in this case.

Double check on viem docs.

x0s0l (changes requested):

src/actions/core/user-positions/common.ts
chimalsky: I will check viem docs about this.

But in this function `getUserPositionData()`... the Promise.all() was wrapping the viem calls. so even if viem uses batch call... this Promise.all( itself will break if one of its functions fail?

bprofiro (approved):
Looks good! thank you for working on it 😄 

I noticed some console.log and console.errors that I don't know if they're intentional or if you were debugging and then forgot about them, so I'm flagging here

src/actions/core/user-positions/common.ts
chimalsky: I actually think what you meant was?

"hey you're not wanting to continue with partial errors here. So just do Promise.all()"

chimalsky: I've pushed `8556c6d0baa3c6583442442a7ca61594f61d560e` to convert to `Promise.all()` and just returning empty array when any of the calls fail

x0s0l (approved):

PR #254, opened 2026-04-09
Fix v2 vaults tvl

bprofiro:
## Fix Morpho V2 vault TVL and historical snapshots                                                                                                                                                                                            
                                                                                                                                                                                                                                               
  Morpho V2 vaults route deposits through a V1 liquidity adapter, so the Morpho API returns only V2's idle assets instead of the actual TVL. APY and rewards are correctly sourced from the indexer.
                                                                                                                                                                                                                                                 
  ### Changes                                                                                                                                                                                                                                    
  - Add `v1VaultKey` to `VaultConfig` to link each V2 vault to its paired V1 vault                                                                                                                                                               
  - Substitute V2 TVL fields (`totalSupply`, `totalLiquidity`, `underlyingPrice`) with V1 data
  - Fetch historical snapshots using the V1 vault address (snapshots are indexed against V1)
  - APY and rewards remain unchanged, using V2's own indexer data

### DIFF
--- a/.changeset/cuddly-carpets-yell.md
+++ b/.changeset/cuddly-carpets-yell.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+For V2 vaults, substitute TVL from the paired V1 vault and use V1 snapshots
--- a/src/actions/morpho/vaults/common.ts
+++ b/src/actions/morpho/vaults/common.ts
@@ -179,6 +179,25 @@ async function getMorphoVaultsDataFromIndexer(params: {
           tokenMap,
         );
 
+        // For V2 vaults, substitute TVL from the paired V1 vault.
+        // V2 routes deposits through V1 via a liquidity adapter, so V1 holds
+        // the actual assets — the Morpho API returns only V2's idle assets (~$13).
+        // APY and rewards are kept from V2's own indexer data.
+        const vaultByKey = new Map(vaults.map((v) => [v.vaultKey, v]));
+        for (const vault of vaults) {
+          const v1VaultKey = environment.config.vaults[vault.vaultKey]
+            ?.v1VaultKey as string | undefined;
+          if (!v1VaultKey) continue;
+          const v1Vault = vaultByKey.get(v1VaultKey);
+          if (!v1Vault) continue;
+          vault.totalSupply = v1Vault.totalSupply;
+          vault.totalSupplyUsd = v1Vault.totalSupplyUsd;
+          vault.vaultSupply = v1Vault.vaultSupply;
+          vault.totalLiquidity = v1Vault.totalLiquidity;
+          vault.totalLiquidityUsd = v1Vault.totalLiquidityUsd;
+          vault.underlyingPrice = v1Vault.underlyingPrice;
+        }
+
         // Filter by specific vault addresses if requested
         if (params.vaults) {
           const requestedVaults = params.vaults.map((id) => id.toLowerCase());
--- a/src/actions/morpho/vaults/common.ts
+++ b/src/actions/morpho/vaults/common.ts
@@ -184,19 +184,28 @@ async function getMorphoVaultsDataFromIndexer(params: {
         // the actual assets — the Morpho API returns only V2's idle assets (~$13).
         // APY and rewards are kept from V2's own indexer data.
         const vaultByKey = new Map(vaults.map((v) => [v.vaultKey, v]));
-        for (const vault of vaults) {
-          const v1VaultKey = environment.config.vaults[vault.vaultKey]
-            ?.v1VaultKey as string | undefined;
-          if (!v1VaultKey) continue;
+        vaults = vaults.map((vault) => {
+          const rawKey = environment.config.vaults[vault.vaultKey]?.v1VaultKey;
+          const v1VaultKey = typeof rawKey === "string" ? rawKey : undefined;
+
+          if (!v1VaultKey) return vault;
+
           const v1Vault = vaultByKey.get(v1VaultKey);
-          if (!v1Vault) continue;
-          vault.totalSupply = v1Vault.totalSupply;
-          vault.totalSupplyUsd = v1Vault.totalSupplyUsd;
-          vault.vaultSupply = v1Vault.vaultSupply;
-          vault.totalLiquidity = v1Vault.totalLiquidity;
-          vault.totalLiquidityUsd = v1Vault.totalLiquidityUsd;
-          vault.underlyingPrice = v1Vault.underlyingPrice;
-        }
+
+          if (!v1Vault) {
+            return vault;
+          }
+
+          return {
+            ...vault,
+            totalSupply: v1Vault.totalSupply,
+            totalSupplyUsd: v1Vault.totalSupplyUsd,
+            vaultSupply: v1Vault.vaultSupply,
+            totalLiquidity: v1Vault.totalLiquidity,
+            totalLiquidityUsd: v1Vault.totalLiquidityUsd,
+            underlyingPrice: v1Vault.underlyingPrice,
+          };
+        });
 
         // Filter by specific vault addresses if requested
         if (params.vaults) {
@@ -474,8 +483,8 @@ async function getMorphoVaultsDataFromOnChain(params: {
         .filter((address) =>
           params.vaults
             ? params.vaults
-                .map((id) => id.toLowerCase())
-                .includes(address.toLowerCase())
+              .map((id) => id.toLowerCase())
+              .includes(address.toLowerCase())
             : true,
         );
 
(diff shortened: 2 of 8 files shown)


apokusin:
# PR #254 Review: Fix V2 Vaults TVL

### Summary

This PR fixes Morpho V2 vault TVL by substituting TVL data from paired V1 vaults, since V2 routes deposits through V1 via a liquidity adapter (causing the Morpho API to return only V2's idle assets ~$13). It adds `v1VaultKey` to `VaultConfig`, applies TVL substitution in the indexer path, redirects snapshot queries to V1 addresses, and extends `typesVersions` in `package.json`. The approach is well-motivated and the comments explain the architecture clearly.

---

### Issues

**🔴 Critical — Snapshot `vaultAddress` returns V1 address instead of requested V2 address**

`src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:56-73`

When a caller requests snapshots for a V2 vault, `vaultAddress` is overwritten with the V1 address before being passed to the fetch functions. Both `fetchVaultSnapshotsFromLunarIndexer` and `fetchVaultSnapshotsFromPonder` embed this (now-V1) address into the returned `MorphoVaultSnapshot.vaultAddress` field. Downstream consumers keying or filtering snapshots by vault address would see a mismatch.

```text
In src/actions/morpho/vaults/getMorphoVaultSnapshots.ts, save the
original V2 vaultAddress before the V1 lookup overwrites it. After
fetching snapshots (both lunarIndexer and ponder paths), map over
the results and replace snapshot.vaultAddress with the original V2
address. For example:
  const originalVaultAddress = vaultAddress;
  // ... V1 lookup and fetch ...
  return snapshots.map(s => ({ ...s, vaultAddress: originalVaultAddress.toLowerCase() }));
```

**🟡 Major — On-chain fallback path has no V1-to-V2 TVL substitution**

`src/actions/morpho/vaults/common.ts` — `getMorphoVaultsDataFromOnChain` function

The V1 TVL substitution was added only to the indexer path. When the indexer is down and the on-chain fallback kicks in, V2 vaults will again report ~$13 TVL — the exact bug this PR is fixing. This silently undermines the fix during indexer outages.

```text
In src/actions/morpho/vaults/common.ts, apply the same V1-to-V2 TVL
substitution logic (the vaultByKey Map + .map() block from the indexer
path) to the getMorphoVaultsDataFromOnChain function. Alternatively,
extract the substitution into a shared helper and call it from both
paths.
```

**🟡 Major — Silent fallbacks with no logging throughout**

`src/actions/morpho/vaults/common.ts:~196` and `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:~68`

Three places where the V1 lookup fails silently, returning incorrect financial data (~$13 TVL or empty snapshots) with no warning:
1. V1 vault not found in the Map despite `v1VaultKey` being configured
2. V1 token address missing in snapshot rewrite
3. Vault-to-key reverse lookup fails entirely

For a financial SDK, silently returning wrong TVL is dangerous. The codebase already uses `console.warn` for similar situations.

```text
In src/actions/morpho/vaults/common.ts around line 197, add a
console.warn when v1Vault is not found despite v1VaultKey being set.
Similarly in src/actions/morpho/vaults/getMorphoVaultSnapshots.ts
around line 70, add a console.warn when v1VaultKey resolved but no
V1 token address was found. Follow the existing console.warn pattern
used elsewhere in common.ts.
```

**🟡 Major — No test coverage for the new logic**

No test files were added or modified. The V1 substitution is the core of this PR and has zero unit test coverage. The snapshot redirect (`getMorphoVaultSnapshots`) has no test file at all. The existing mock patterns in `lunarIndexerTransform.test.ts` and `getMorphoVaultStakingSnapshots.test.ts` make adding tests straightforward.

Key cases to cover:
- V2 vault with valid `v1VaultKey` → TVL fields copied from V1, APY/rewards preserved
- V2 vault with `v1VaultKey` but V1 vault missing → original vault returned
- V1 vault or vault without `v1VaultKey` → unchanged
- Snapshot address redirect for V2 vs V1 vaults

**🔵 Minor — Duplicated `v1VaultKey` resolution logic**

`src/actions/morpho/vaults/common.ts:~189` and `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:~60-67`

Both files independently implement the same `typeof rawKey === "string"` narrowing pattern. A shared helper like `getV1VaultKey(environment, vaultKey): string | undefined` would centralize this and eliminate duplication.

**🔵 Minor — `let` destructuring broader than needed**

`src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:47`

The `const` was changed to `let` for the entire destructuring, but only `vaultAddress` is reassigned. Cleaner to destructure as `const` and declare `let vaultAddress` separately.

**🔵 Minor — Reverse lookup uses vault key as token key by convention**

`src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:~61`

The snapshot code uses `environment.config.tokens[vaultKey]?.address` assuming vault keys equal token keys. The vault config has a `vaultToken` field that explicitly names the token key — using `vaultConfig.vaultToken` would be more robust.

---

### Positive Observations

- Well-written PR description clearly explaining the V2-wraps-V1 architecture
- Correctly preserves V2's APY and reward data while only substituting TVL fields
- `v1VaultKey` typed as `keyof tokens` provides compile-time safety at definition sites
- Good use of optional chaining throughout, consistent with project conventions (no `!` or `as`)
- Efficient `Map`-based O(1) lookup for V1 vault resolution
- `typesVersions` additions correctly mirror the existing `exports` entries

---

### Suggestions

- **Extract V1 substitution into a shared helper** — both for reuse across indexer/on-chain paths and to eliminate the duplicated type narrowing
- **Consider a discriminated union for `VaultConfig`** — `VaultV1Config` vs `VaultV2Config` (with required `v1VaultKey`) would prevent invalid states like `version: 2` without a paired V1 key. However, the `Environment` type erases the generic, so this mainly helps at the definition sites in `morpho-vaults.ts`
- **Add a runtime check that the V1 target isn't itself a V2 vault** — prevents circular or nonsensical pairings

---

### Recommended Action

1. **Fix** the snapshot `vaultAddress` mismatch bug (🔴)
2. **Add** V1 substitution to the on-chain fallback path, or at minimum log a warning (🟡)
3. **Add** `console.warn` logging on silent fallback paths (🟡)
4. **Add** unit tests for the substitution and snapshot redirect logic (🟡)
5. Consider the minor suggestions at your discretion

apokusin:
Reviewing against `main`, I found 3 issues:

1. **[P1] `src/actions/morpho/vaults/common.ts:182-207` — V1 TVL swap leaves V2 market breakdown on a different ownership model**

   This copies `totalSupply` / `totalLiquidity` / `vaultSupply` from the paired V1 vault but leaves the V2 `markets` array untouched. That makes the returned `MorphoVault` internally inconsistent: the top-line TVL is now describing the V1 vault, while the per-market breakdown is still describing the V2 vault.

   I verified this against the live Lunar payloads on Base. For `mwUSDC`, the V2 vault currently has about `39.4` assets, while `mwUSDCv1` has about `13.93M`. After this change, `mwUSDC.totalSupply` would jump to the V1 number, but `mwUSDC.markets` would still sum to about `39.4`. Consumers can no longer reconcile headline TVL with the market allocation breakdown.

   The fix needs to source the market fields from the same V1 series as the totals, or avoid mixing V1 totals with V2 markets.

2. **[P1] `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:53-86` — Snapshot responses are now keyed to the wrong vault address**

   This rewrites `vaultAddress` to the paired V1 vault before both fetch paths, and the returned snapshots preserve that queried address. A caller asking for a V2 vault now gets snapshots whose `vaultAddress` is the V1 vault, which breaks the API contract for any caller that joins or caches snapshots by the requested vault address.

   Using V1 as the data source is reasonable, but the response should be normalized back to the original V2 address.

3. **[P2] `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:53-86` — V1 snapshot remap leaks pre-launch history into V2 charts**

   The remap is unconditional and there is no clipping by the V2 vault's own inception time. On live data, `meUSDCv1` started on **September 25, 2025**, while `meUSDC` started on **December 10, 2025**. With this change, `period: "ALL"` or broad custom ranges will now return history from before the V2 vault existed.

   If V1 snapshots are presented as V2 history, they need to be truncated at the V2 launch timestamp.

I wasn’t able to run the Vitest suite locally in this worktree because `node_modules` is absent here, so `pnpm exec vitest` fails immediately with `Command "vitest" not found`. Otherwise the review is based on the PR diff plus the current live Lunar payloads.

apokusin (dismissed):

apokusin:
# PR #254 Revised Review: Fix V2 Vaults TVL

> Revised with context: V1/V2 are the same vault from the user's perspective. The V2-routes-through-V1 split is an implementation detail. Blending TVL from V1 with APY/rewards from V2 is the intended design.

### Summary

Clean approach — adds a declarative `v1VaultKey` config field to pair V2 vaults with their V1 counterpart, substitutes TVL fields in the indexer path, and redirects snapshot queries to V1 addresses. The blending logic is well-scoped (only TVL fields, preserving V2 APY/rewards) and the config-driven design keeps business logic out of the data layer.

---

### Issues

**🟡 Major — Add `console.warn` when V1 vault lookup fails**

`src/actions/morpho/vaults/common.ts:~195` and `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:~68`

Falling back to V2 data when V1 isn't found is fine as degradation — but the codebase already uses `console.warn` extensively in `common.ts` for analogous fallback paths (lines 203, 226, 266, 401). Two places should match that pattern:

1. `common.ts` — when `v1VaultKey` is set but V1 vault isn't in the indexer response
2. `getMorphoVaultSnapshots.ts` — when `v1VaultKey` resolves but no token address is found

Without these, a config typo or indexer omission causes V2 to silently display ~$13 TVL with zero log evidence.

```text
In src/actions/morpho/vaults/common.ts around line 195, add a console.warn
before `return vault` inside the `if (!v1Vault)` guard. Include both the V2
vault key and the expected V1 key. Follow the style at line 266.

In src/actions/morpho/vaults/getMorphoVaultSnapshots.ts around line 68, add
an else branch to `if (v1Token?.address)` with a console.warn including the
v1VaultKey value.
```

---

### Positive Observations

- Config-driven `v1VaultKey` approach is clean and declarative — keeps the V1/V2 pairing in environment definitions
- TVL substitution correctly preserves V2 APY, rewards, and market data
- `keyof tokens` typing catches config key typos at compile time
- Defensive null-safety throughout (optional chaining, Map.get guards) — no crash risks
- Efficient O(1) Map-based V1 lookup
- Snapshot address redirect correctly serves both Lunar Indexer and Ponder paths without duplication
- `typesVersions` additions correctly mirror existing `exports` entries

---

### Suggestions

**On-chain fallback path** — `getMorphoVaultsDataFromOnChain` does not include the V1 substitution. During indexer outages V2 vaults will revert to ~$13 TVL. Worth considering whether to add the same substitution there (V1 on-chain data is already queried), or accept it as a known limitation of the degraded path.

**`let` destructuring scope** — In `getMorphoVaultSnapshots.ts:47`, only `vaultAddress` is reassigned but the entire destructuring was changed to `let`. Cleaner to destructure as `const` and declare `let vaultAddress` separately.

**`typeof rawKey === "string"` simplification** — The check in both files guards against `number | symbol` keys that can't actually occur at runtime (the erased type is `string | undefined`). Non-blocking, but could be simplified to just `const v1VaultKey = vaultConfig.v1VaultKey; if (!v1VaultKey) return vault;`.

**Consider hiding V1 vaults from public results** — Both V1 and V2 vaults are returned to callers. Since they represent the same vault, consumers may see duplicates. Could filter V1 vaults out after substitution if they shouldn't be user-visible.

---

### Recommended Action

1. **Add** `console.warn` on fallback paths (🟡) — debuggability, matches codebase convention
2. Consider suggestions at your discretion

apokusin (approved):

PR #262, opened 2026-04-20
Remove ponder/subgraph fallback from moonwell sdk rely on

bprofiro:
## Summary
                                                                                                                                                                                                                                             
  Removes all Ponder/subgraph (Goldsky) fallback code from the SDK and introduces resilience improvements for the Lunar Indexer as the primary data source.                                                                                      
                                                                                                                                                                                                                                                 
  ### Ponder / subgraph removal                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                 
  - Removes `indexerUrl` from all environment configs and the `Environment` type (breaking change)
  - Removes the following dead functions:
    - `fetchCoreMarketSnapshotsFromPonder`
    - `fetchUserPositionSnapshotsFromPonder`
    - `fetchStakingSnapshotsFromPonder`
    - `fetchMorphoGraphQL`
    - `fetchIsolatedMarketSnapshotsFromBlueApi`
  - Snapshot functions (`getMarketSnapshots`, `getStakingSnapshots`, `getUserPositionSnapshots`, `getCirculatingSupplySnapshots`) now return `[]` when `lunarIndexerUrl` is absent instead of falling back to Ponder
  - **Exception:** Moonriver governance proposals retain their Ponder-based path (`getExtendedProposalData`) to preserve proposal descriptions on moonwell.fi/governance

  ### On-chain RPC fallback for live data

  - `getMarketsData` now tries Lunar Indexer first and falls back to on-chain RPC if it fails (previously failed silently or threw)
  - `getMorphoVaultsData` already had an RPC fallback; now consistently triggered by any Lunar failure

  ### `onError` callback

  Adds an optional `onError` callback to `createMoonwellClient` for error observability (e.g. Sentry logging). It is called whenever Lunar Indexer fails, even when the SDK recovers via fallback.

  ```ts
  const client = createMoonwellClient({
    networks: { ... },
    onError: (error, { source, chainId }) => {
      Sentry.captureException(error, { tags: { source, chainId } });
    },
  });

  The callback receives a source string ("markets" or "vaults") and the chainId of the failing environment.

  lunarIndexerUrl consolidation

  - Removes lunarIndexerUrl from custom.morpho config in Base and Optimism environments (breaking change for anyone setting this field directly)
  - All Lunar Indexer access (markets, vaults, snapshots, user positions) now uses the single top-level environment.lunarIndexerUrl

  Tests

  - src/actions/core/markets/common.test.ts — covers Lunar routing, onError invocation, and RPC fallback for getMarketsData
  - src/actions/morpho/vaults/getMorphoVaultsData.test.ts — covers lunarIndexerUrl routing, onError invocation, and fallback behaviour for getMorphoVaultsData
  - src/client/createMoonwellClient.test.ts — covers onError wiring onto all created environments


apokusin:
## Critical Issues from Review

### 🔴 1. Morpho vault snapshot actions miss `onError` + try/catch
`getMorphoVaultSnapshots`, `getMorphoVaultStakingSnapshots`, and `getMorphoVaultUserPositionSnapshots` call their Lunar fetchers with no error handling. A Lunar failure throws to the caller instead of routing through `onError` + `[]` like every other snapshot action.
- `src/actions/morpho/vaults/getMorphoVaultSnapshots.ts:83-101`
- `src/actions/morpho/vaults/getMorphoVaultStakingSnapshots.ts:61-68`
- `src/actions/morpho/user-positions/getMorphoVaultUserPositionSnapshots.ts:58-69`

```text
In src/actions/morpho/vaults/getMorphoVaultSnapshots.ts (~line 83), src/actions/morpho/vaults/getMorphoVaultStakingSnapshots.ts (~line 61), and src/actions/morpho/user-positions/getMorphoVaultUserPositionSnapshots.ts (~line 64), wrap the call to the lunar fetch helper in a try/catch. On error, log via console.warn, call environment.onError?.(error, { source: "morpho-vault-snapshots" | "morpho-vault-staking-snapshots" | "morpho-vault-user-position-snapshots", chainId: environment.chainId }), and return []. This matches the pattern already established in getMarketSnapshots, getStakingSnapshots, and getCirculatingSupplySnapshots.
```

### 🔴 2. `getMorphoMarketsDataFromIndexer` doesn't invoke `onError` and leaves RPC fallback uncaught
`src/actions/morpho/markets/common.ts:805-858` logs with `console.warn` only (no `onError`), and the subsequent `getMorphoMarketsDataFromOnChain` fallback isn't wrapped — if RPC fails the entire call rejects and the original Lunar failure context is lost. Also uses `(s: any)` to read rejection reasons.

```text
In src/actions/morpho/markets/common.ts around lines 826-858, inside the Promise.allSettled catch block call environment.onError?.(error, { source: "morpho-markets", chainId: environment.chainId }) before rejecting. Wrap getMorphoMarketsDataFromOnChain at line 853 in a try/catch — on fallback failure, call environment.onError with source "morpho-markets-rpc-fallback" for each failed environment and return [] rather than letting the error propagate. Remove the `s: any` cast by typing the rejection reason properly.
```

### 🔴 3. Retained Ponder paths have zero error handling
`fetchCoreMarketSnapshotsFromPonder`, `fetchUserPositionSnapshotsFromPonder`, `getProposalData`, `getCrossChainProposalData`, and `getExtendedProposalData` all perform axios/RPC calls with no try/catch — the Moonriver fallback is strictly worse than the Lunar path it replaces. `getExtendedProposalData` also lacks an iteration cap.

```text
In src/actions/core/markets/getMarketSnapshots.ts (~line 143-167) and src/actions/core/user-positions/getUserPositionSnapshots.ts (~line 91-95), wrap the Ponder fallback calls in try/catch that console.warn, invoke environment.onError with source "market-snapshots-ponder" / "user-position-snapshots-ponder" and chainId, and return [].

In src/actions/governance/proposals/common.ts, wrap getProposalData (320-409), getCrossChainProposalData (414-588), and getExtendedProposalData (616-711) in try/catch. On failure call environment.onError?.(error, { source: "governance-proposals", chainId: environment.chainId }) and return []. Add a MAX_PAGES = 100 cap to the while loop in getExtendedProposalData.
```

### 🔴 4. Non-null assertion + `as` cast in `createMoonwellClient`
`src/client/createMoonwellClient.ts:87-89` uses `(config.networks as NetworksConfig<SupportedChains>)[key]!` — violates the "no `!`, no `as`" project convention in newly touched code.

```text
In src/client/createMoonwellClient.ts around line 87-89, remove the `as NetworksConfig<SupportedChains>` type assertion and the non-null `!`. Restructure the reduce so Object.entries(config.networks) yields typed pairs and narrow with an if-check that the value is defined. Preserve runtime behavior exactly.
```

### 🔴 5. `getMarketsData` RPC fallback uses `.value!` on `Promise.allSettled` results
`src/actions/core/markets/common.ts:69-91`. Fulfilled values can still be `undefined` (e.g. `viewsContract?.read.getProtocolInfo()`), producing opaque TypeErrors that bypass `onError`.

```text
In src/actions/core/markets/common.ts around lines 69-91, replace every `value!` on protocolInfoResult, allMarketsInfoResult, nativePriceResult, and govPriceResult with explicit guards that handle the "fulfilled but undefined" case. When a required value is undefined after a fulfilled settlement, call environment.onError with source "markets-onchain-missing-views" and return [] rather than throwing an opaque TypeError.
```

apokusin (dismissed):
version bump presumably handled via CI?
might be worth running `/simplify` to clean up syntax/comments - optional

avp1598 (approved):

PR #266, opened 2026-05-04
fix per-chain snapshot block across multi-chian governance tokens

bprofiro:
## Summary                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                 
  Adds `snapshotTimestamp` to `getUserVotingPowers`. When set, the SDK resolves the block on each queried chain whose timestamp is the latest one ≤ `snapshotTimestamp`, then reads voting power at that per-chain block.                        
                                                                                                                                                                                                                                                 
  This fixes [MOO-246](https://linear.app/moonwell/issue/MOO-246): for cross-chain governance tokens (WELL spans Moonbeam, Base, Optimism), forwarding a single Moonbeam block height to all chains hit the Optimism views contract at a 2022-era block (before deployment) and returned `0x`. Resolving per chain via timestamp avoids this entirely.                                                                                                                                          
                                                                                                                                                                                                                                                 
  The legacy `blockNumber` parameter still works (applied uniformly to every chain) — fine for single-chain tokens like MFAM.                                                                                                                    
   
  **Changes:**                                                                                                                                                                                                                                   
  - New helper `common/getBlockNumberAtTimestamp` — interpolation search anchored on the latest block + block 1, ~3–5 RPC calls per chain even with variable block times
  - `getUserVotingPowers` accepts `snapshotTimestamp?: number`; when set, resolves per-chain blocks before contract reads                                                                                                                        
                                                                                                                                                                                                                                                 
  **Public API:** additive — `snapshotTimestamp` is optional, no existing call sites break. Patch bump.                                                                                                                                          
                                                                                                                                                                                                                                                 
  ## Test plan                                                                                                                                                                                                                                   
                                                            
  - [x] `pnpm typecheck` passes
  - [x] `pnpm lint` passes
  - [x] `pnpm build` produces `_esm/common/getBlockNumberAtTimestamp.js`


apokusin:
# Review

### Summary
Adds `snapshotTimestamp` to `getUserVotingPowers` so the SDK resolves a per-chain block whose timestamp is the latest one ≤ target before reading on-chain voting power. New helper `getBlockNumberAtTimestamp` does interpolation search anchored on `latest` and block 1 (~3–5 RPC calls). Also collapses the duplicated mapped-type expression in `config.ts` into a single named helper. Net change is small (+135 / −216).

### Issues

#### 🟡 Major: No test coverage for the new helper or the new parameter
`src/common/getBlockNumberAtTimestamp.ts` has no unit tests, and `getUserVotingPowers.test.ts` is not updated to exercise `snapshotTimestamp`. This PR specifically fixes a bug; without a regression test the same class of bug can re-emerge silently. The interpolation routine has several edge cases worth pinning (target ≥ latest, target ≤ block 1, single-block chain, range that converges in 1 iteration, target equal to a block timestamp).

```text
In src/common/getBlockNumberAtTimestamp.ts, add a co-located test file (getBlockNumberAtTimestamp.test.ts) that mocks a viem PublicClient's getBlock and covers: (a) target ≥ latest.timestamp returns latest.number, (b) target ≤ block-1.timestamp returns 1n, (c) target between two known block timestamps returns the largest block whose timestamp is ≤ target, (d) convergence within ≤ 8 iterations on a fabricated 1M-block chain, (e) variable block-time chain (alternating 2s and 30s gaps). Use a Map<bigint, {number,timestamp}> as the fixture so the mock is deterministic.
```

```text
In src/actions/governance/getUserVotingPowers.test.ts, add a test that passes snapshotTimestamp and asserts views.read.getUserVotingPower is invoked with a per-chain blockNumber resolved from getBlockNumberAtTimestamp (mock the helper). Also add a test that confirms snapshotTimestamp takes precedence over blockNumber when both are supplied.
```

#### 🟡 Major: Type assertion violates project conventions
`src/actions/governance/getUserVotingPowers.ts` casts `env.publicClient as PublicClient` inside the `Promise.all` over `tokenEnvironments`. `config.ts` already declares `publicClient: PublicClient`, so the cast is either redundant (drop it) or hiding a real type mismatch (fix it upstream). CLAUDE.md explicitly forbids `as Type` in non-test code.

```text
In src/actions/governance/getUserVotingPowers.ts, remove the `as PublicClient` cast on env.publicClient inside the Promise.all over tokenEnvironments. If TypeScript still complains, investigate why — likely getEnvironmentsFromArgs returns a union that loses the PublicClient annotation; in that case fix the upstream type rather than casting at the call site.
```

#### 🔵 Minor: `getBlockNumberAtTimestamp` silently caps at 8 iterations
The hardcoded `for (let i = 0; i < 8; i += 1)` will silently return a possibly-loose `lo` if interpolation fails to converge (highly skewed block-time distribution, reorgs between calls). For governance snapshots the looseness is acceptable, but the constant deserves a name and a comment.

```text
In src/common/getBlockNumberAtTimestamp.ts, extract the magic number 8 into a named const (e.g. MAX_ITERATIONS) at the top of the file with a short comment explaining the convergence reasoning (interpolation search on near-linear ts(block) converges in ~3–5 iterations; 8 is a safety margin for variable block times).
```

#### 🔵 Minor: Edge case — chain with only block 0
If `latest.number === 0n` and `targetTimestamp < latest.timestamp`, the function falls through and returns `lo = 0n` with `loTs = 0n` — claiming block 0 has timestamp 0, which is false. No production chain hits this, but it's a small correctness gap.

```text
In src/common/getBlockNumberAtTimestamp.ts, handle latest.number === 0n explicitly: return latest.number directly since there is no earlier block to resolve to. This removes the pseudo-initialization loTs = 0n which is technically inaccurate.
```

#### 🔵 Minor: `getBlock({ blockNumber: 1n })` assumes block 1 exists
True for every EVM chain in the Moonwell footprint, but worth a sentence in the JSDoc.

### Positive observations
- The `EnvironmentContracts<contracts>` consolidation drops ~200 lines of nearly-identical mapped-type expressions while preserving semantics.
- JSDoc on `snapshotTimestamp` is clear about the precedence rule and *why* a single block number across chains is unsafe.
- Interpolation search is the right primitive for this problem and the implementation is tight (~30 lines).
- Public API change is genuinely additive; patch bump and changeset entry are appropriate.
- Per-chain `getBlockNumberAtTimestamp` calls are parallelized via `Promise.all`.

### Suggestions
Consider memoizing per-`(chainId, timestamp)` results inside the client if `getUserVotingPowers` is called repeatedly for the same snapshot (governance UIs often do this). Out of scope here but worth a follow-up.

```text
Open a follow-up issue to investigate caching getBlockNumberAtTimestamp results keyed by (chainId, timestamp) at the MoonwellClient level, since governance dashboards commonly resolve the same snapshot timestamp many times across users/proposals.
```

### Recommended action
1. Add tests for `getBlockNumberAtTimestamp` and `getUserVotingPowers(snapshotTimestamp)` — main blocker.
2. Remove the `as PublicClient` cast (or fix the upstream type if load-bearing).
3. Address the minor items at your discretion before merge.

apokusin (approved):

PR #272, opened 2026-05-15
feat(morpho/user-rewards): opt-in error propagation for Merkl API

bprofiro:
## Summary

- Adds an optional `throwOnExternalApiError` parameter to `getMorphoUserRewards`. When `true`, Merkl API failures propagate to the caller instead of being silently swallowed and returning `[]`.
- Default remains `false` — every existing consumer sees no behavior change.
- Pairs with frontend graceful-degradation work: the rewards-page hook opts in so React Query can surface an inline "Morpho rewards unavailable" notice when Merkl is down, instead of indistinguishably showing "no Morpho rewards".

## Test plan

- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm lint` clean (Biome)
- [x] `pnpm test src/actions/morpho/user-rewards/common.test.ts` — 7/7 pass, including two new cases covering throw-on-error for non-ok responses and fetch rejections
- [x] `pnpm build` produces _cjs / _esm / _types successfully
- [ ] Companion frontend PR review (separate)

### DIFF
--- a/.changeset/tall-lights-write.md
+++ b/.changeset/tall-lights-write.md
@@ -1,5 +1,9 @@
 ---
-"@moonwell-fi/moonwell-sdk": patch
+"@moonwell-fi/moonwell-sdk": minor
 ---
 
-Add opt-in `throwOnExternalApiError` option to `getMorphoUserRewards`. When `true`, Merkl API failures propagate to the caller instead of being swallowed and returning `[]`. Defaults to `false`, so existing consumers see no behavior change.
+Add opt-in `throwOnExternalApiError` option to `getMorphoUserRewards`. When `true`, Merkl API failures (non-ok HTTP responses, network rejections, and response-body parse errors) propagate to the caller as `MerklApiError` instead of being swallowed and returning `[]`. Defaults to `false`, so existing consumers see no behavior change.
+
+When multiple chains are queried and at least one fails while others succeed, the action throws a `MorphoUserRewardsAggregateError` whose `errors` array carries the per-chain failures and whose `rewards` property carries the rewards from chains that succeeded, so callers can surface partial results alongside the failures.
+
+New public exports: `MerklApiError`, `MorphoUserRewardsAggregateError`.
--- a/.changeset/tall-lights-write.md
+++ b/.changeset/tall-lights-write.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Add opt-in `throwOnExternalApiError` option to `getMorphoUserRewards`. When `true`, Merkl API failures propagate to the caller instead of being swallowed and returning `[]`. Defaults to `false`, so existing consumers see no behavior change.
--- a/src/actions/morpho/user-rewards/common.test.ts
+++ b/src/actions/morpho/user-rewards/common.test.ts
@@ -174,6 +174,43 @@ describe("getUserMorphoRewardsData", () => {
     expect(result).toEqual([]);
   });
 
+  test("with throwOnExternalApiError, propagates non-ok Merkl responses", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(() =>
+        Promise.resolve({
+          ok: false,
+          status: 500,
+          statusText: "Internal Server Error",
+        }),
+      ),
+    );
+
+    await expect(
+      getUserMorphoRewardsData({
+        environment: baseEnvironment,
+        account: ACCOUNT,
+        throwOnExternalApiError: true,
+      }),
+    ).rejects.toThrow(/Merkl API request failed: 500/);
+  });
+
+  test("with throwOnExternalApiError, propagates fetch rejections", async () => {
+    const networkError = new Error("network unreachable");
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(() => Promise.reject(networkError)),
+    );
+
+    await expect(
+      getUserMorphoRewardsData({
+        environment: baseEnvironment,
+        account: ACCOUNT,
+        throwOnExternalApiError: true,
+      }),
+    ).rejects.toBe(networkError);
+  });
+
   test("on full deployments, returns zero claimable when no breakdowns match Moonwell campaigns", async () => {
     mockFetchOnce(
       makeMerklResponse({ amount: "1000", claimed: "200", pending: "50" }, [
--- a/src/actions/morpho/user-rewards/common.test.ts
+++ b/src/actions/morpho/user-rewards/common.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, describe, expect, test, vi } from "vitest";
 import type { Environment } from "../../../environments/index.js";
 import type { MorphoUserReward } from "../../../types/morphoUserReward.js";
-import { getUserMorphoRewardsData } from "./common.js";
+import { MerklApiError, getUserMorphoRewardsData } from "./common.js";
 
 type MerklReward = Extract<MorphoUserReward, { type: "merkl-reward" }>;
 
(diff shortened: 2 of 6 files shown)


apokusin:
## Summary

This PR adds an opt-in `throwOnExternalApiError` parameter to `getMorphoUserRewards`. When set, Merkl API failures (non-ok HTTP responses, network rejections, JSON parse errors) propagate to the caller instead of being silently swallowed and returning `[]`. The default behavior is preserved, so existing consumers see no change. A new `MerklApiError` class captures HTTP-failure context (status, statusText, url, chainId). Multi-chain failures are surfaced as an `AggregateError`. A changeset is included (patch).

The public API change is additive (a new optional parameter and a new exported error class). Tests cover both code paths (flag-on / flag-off, HTTP-error vs network-rejection vs success, single chain vs multi-chain). The new test file is picked up by the existing `src/actions/morpho/**/*.test.ts` glob in `test/vitest.config.ts:40`, so no `include` update is needed.

## Issues

### 🟡 Major

**`MerklApiError` is exported but not re-exported on a public entry point** — `src/actions/morpho/user-rewards/common.ts:16`

The class is exported from `common.ts` but `common.ts` is not part of the published surface, so consumers can't do `import { MerklApiError } from "@moonwell-fi/moonwell-sdk"` to perform `instanceof` checks. The PR description mentions the frontend will surface a "Morpho rewards unavailable" notice — if it wants to distinguish HTTP failures from network failures, it currently has to inspect `aggregate.errors[0].cause` and string-match the message. Either re-export `MerklApiError` from the SDK's public entry point, or document that the only contract consumers should rely on is "rejection means external failure" (in which case the class can stay internal).

```text
In src/actions/morpho/user-rewards/common.ts the `MerklApiError` class is exported but not re-exported through a public entry point. Verify which file is the public re-export surface for getMorphoUserRewards (likely src/actions/morpho/user-rewards/index.ts or src/index.ts) and add `export { MerklApiError } from "./common.js";` there so consumers can do `instanceof MerklApiError` checks. If we don't want it to be part of the public surface, remove the export and adjust the changeset and types accordingly.
```

**`throwOnExternalApiError: true` discards rewards from successful chains** — `src/actions/morpho/user-rewards/getMorphoUserRewards.ts:78-83`

When the flag is on and one of N chains fails, the entire result is thrown away — `fulfilled` is built but never returned in the error path. The JSDoc warns about this, but it's a sharp edge: a frontend showing "Morpho rewards unavailable" because Base 5xx'd would lose Optimism rewards the user actually has. Two reasonable options: (a) attach `fulfilled` as a property on the `AggregateError` so callers can opt into partial success, or (b) document explicitly that single-chain consumers are the intended use and multi-chain callers should set `throwOnExternalApiError: false` and inspect logs.

```text
In src/actions/morpho/user-rewards/getMorphoUserRewards.ts (the `if (failures.length > 0 && args.throwOnExternalApiError === true)` block around line 78), attach the partial successes to the AggregateError so consumers can surface them. Subclass AggregateError (e.g. `class MorphoUserRewardsAggregateError extends AggregateError { rewards: MorphoUserReward[] }`) or use Object.assign to attach a `rewards` property to the thrown AggregateError. Update the JSDoc on GetMorphoUserRewardsParameters.throwOnExternalApiError to describe the new behavior. Add a vitest case to getMorphoUserRewards.test.ts that asserts the thrown AggregateError carries the successful-chain rewards alongside the failures.
```

### 🔵 Minor

**Double error wrapping loses `MerklApiError` type for HTTP failures** — `src/actions/morpho/user-rewards/getMorphoUserRewards.ts:64-72`

For non-ok HTTP responses, `common.ts` throws a `MerklApiError`, which `getMorphoUserRewards.ts` then wraps in a plain `Error` (with `cause`), which then goes into the `AggregateError`. Consumers wanting the HTTP status now have to do `aggregate.errors[0].cause` and downcast. The wrapping adds chainId to the message, but that's already on `MerklApiError.chainId`. Consider pushing the original error directly into the AggregateError without an intermediate `new Error` layer when the underlying error is already typed (or when it already carries chain context).

```text
In src/actions/morpho/user-rewards/getMorphoUserRewards.ts (lines 64-72), avoid double-wrapping errors that already carry chain context. If `result.reason instanceof MerklApiError`, push it directly into `failures` without wrapping in a plain Error. For other errors (raw network/parse failures from common.ts that aren't MerklApiError), keep the current wrapping so the chain is still attached. Update the matching getMorphoUserRewards.test.ts assertion so `aggregate.errors[0]` is the MerklApiError itself when the inner error was a MerklApiError.
```

**Inconsistent inner-error types** — `src/actions/morpho/user-rewards/common.ts:340-371`

The non-ok HTTP branch throws a typed `MerklApiError`, but the network branch (`fetch` rejection) and the parse branch (`response.json()` rejection) re-throw the raw error untyped. So `aggregate.errors[0].cause` is `MerklApiError` for HTTP failures and a plain `Error` / `SyntaxError` for network or parse failures. Either wrap all three in `MerklApiError` with different `status: undefined`, or document the variance in `MerklApiError`'s JSDoc.

```text
In src/actions/morpho/user-rewards/common.ts within getMerklRewardsData, wrap the network-rejection and parse-rejection branches in MerklApiError as well (with status/statusText undefined) so the error shape is uniform. The constructor already accepts optional status/statusText. Set message to something distinguishing the cause (e.g. "Merkl API network error for chain N" / "Merkl API response parse error for chain N") and chain the original via `{ cause: error }` if you extend MerklApiError to accept it. Update common.test.ts to assert MerklApiError is thrown on fetch rejection too when throwOnExternalApiError is set.
```

**Dead-code defensive check on `targetEnvironments[i]`** — `src/actions/morpho/user-rewards/getMorphoUserRewards.ts:60-62`

`Promise.allSettled` returns an array whose length is exactly the input array's length, so `settled[i]` and `targetEnvironments[i]` are always defined within the loop. The `if (result === undefined || environment === undefined) continue;` guard exists only to silence `noUncheckedIndexedAccess`. Consider iterating with `for (const [i, result] of settled.entries())` and destructuring; the environment lookup can stay indexed but the guard can be dropped if we accept a `// biome-ignore` comment, or kept as-is — flagging only because CLAUDE.md discourages defensive code for impossible scenarios.

**Redundant conditional spread** — `src/actions/morpho/user-rewards/getMorphoUserRewards.ts:43-45`

`...(args.throwOnExternalApiError !== undefined && { throwOnExternalApiError: args.throwOnExternalApiError })` reads awkwardly. If tsconfig has `exactOptionalPropertyTypes`, this matters; otherwise `throwOnExternalApiError: args.throwOnExternalApiError` works identically and is clearer. Worth a quick check against the project's tsconfig — if `exactOptionalPropertyTypes` is off, simplify.

**Changeset bump level** — `.changeset/tall-lights-write.md`

A `patch` is reasonable since the new parameter is opt-in and behavior is preserved, but a new public export (`MerklApiError`) and a new public option are typically `minor` in semver. Up to the maintainers, but flagging for awareness.

## Positive observations

- **Backwards compatibility is well-preserved.** Default behavior unchanged, opt-in flag, and the existing test case `returns empty array when fetch rejects and throwOnExternalApiError is not set` explicitly locks in the old behavior.
- **Good test coverage of the new branches** — both single-chain and multi-chain failure modes, both flag states, and shape assertions on `AggregateError.errors[i].cause`.
- **Logging improvements are nice.** The previous warn/error lines lacked chain context and URL; the new format (`chain=X url=Y`) makes triage materially easier even for callers who stay on the default behavior.
- **Refactor of `getMerklRewardsData` from one big try/catch to discrete try blocks** correctly separates network failures, HTTP failures, and parse failures — three distinct categories that previously all logged as the same generic "Error in getMerklRewardsData".
- **No production-code `as` casts or non-null assertions added.** The casts in the test files are within the CLAUDE.md exception.
- **JSDoc on `throwOnExternalApiError`** explicitly warns about the all-or-nothing behavior on multi-chain failures, even though that warning belongs in code (see Major #2).

## Suggestions

**Add a test asserting `console.warn`/`console.error` are NOT called when `throwOnExternalApiError: true`.** The current swallow path logs, but the throw path should be log-silent (the caller will log). This guarantees we don't double-report.

```text
In src/actions/morpho/user-rewards/common.test.ts, add tests that spy on console.warn and console.error and assert they are not called when getUserMorphoRewardsData throws under throwOnExternalApiError: true (for both the non-ok HTTP case and the fetch-rejection case). Use vi.spyOn(console, "warn"/"error") and assert .toHaveBeenCalledTimes(0).
```

**Consider exporting `getMerklRewardsData` for direct use** if frontends sometimes want only Merkl data without the Moonwell-vault filtering wrapper. Not in scope for this PR, but worth a follow-up issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

apokusin (approved):

PR #276, opened 2026-05-23
fix(governance): route WELL price through Base oracle; stkWELL fallback for Moonbeam

lyoungblood:
## Summary

The per-chain `views.getGovernanceTokenPrice()` is unreliable. Verified via direct `eth_call`:

| Call | Returns | Verdict |
|---|---|---|
| Moonbeam `getGovernanceTokenPrice()` | `0xb51a461234c478` → ~\$0.0510 | Stale / wrong |
| Base `getGovernanceTokenPrice()` | `0x0` | Broken (returns 0) |
| Base `oracle.getUnderlyingPrice(mWELL)` | `0xdc590b501e000` → ~\$0.0038764 | **Correct (Chainlink-fed)** |
| Moonbeam `views.getStakingInfo()` | reverts | Why `/stake/moonbeam` hangs |
| Moonbeam `views.getUserStakingInfo(addr)` | reverts | Why Moonbeam users see no stake data |

Before this PR, the `homeEnvironment` lookup routed every WELL-governed chain's price read through Moonbeam's broken views contract — so `/stake/base` displayed Moonbeam's bogus \$0.0522 instead of the real ~\$0.0039.

### `getGovernanceTokenPriceFor(env, baseEnv?)` — dispatcher (`src/actions/governance/getWellPrice.ts`)
- If `env.custom.governance.token === \"WELL\"` → reads `oracle.getUnderlyingPrice(mWELL)` on Base.
- Else (Moonriver / MFAM) → reads the env's own `views.getGovernanceTokenPrice()`. Moonriver was previously treated correctly only by accident because no env's `chainIds` included 1285; this PR makes the dispatch explicit so it stays correct under refactors.
- `baseEnvironment` is forwarded from the caller's `MoonwellClient` so user-configured Base RPC and `onError` handlers are honored.
- Fires `baseEnv.onError` if Base env is missing `MOONWELL_WELL` token or oracle contract (was silently returning 0n).

### Wired through every governance-price call site
- `governance/getStakingInfo.ts`, `governance/getUserStakingInfo.ts`
- `core/markets/common.ts` (2 sites), `core/user-rewards/common.ts`
- `morpho/user-rewards/common.ts`, `morpho/vaults/common.ts` (2 sites)

At each `Promise.all` site, the call is wrapped in `.catch((err) => { env.onError?.(err, ...); return 0n; })` so a Base RPC blip doesn't cascade into non-Base flows AND the failure is observable.

### Moonbeam staking fallback — `src/actions/governance/getStakingInfo.ts`
`views.getStakingInfo()` reverts on Moonbeam. New `getStakingInfoFromStkWell` reads the same fields directly from the stkWELL contract:
- `COOLDOWN_SECONDS()`, `UNSTAKE_WINDOW()`, `DISTRIBUTION_END()`, `totalSupply()`
- `assets([stakingToken.address])[0]` for `emissionPerSecond` — keyed by the stkWELL contract's own address per Aave-fork convention (`address(this)` in `StakedAave._initialize`). Verified on Moonbeam: `assets(stkWELL)` returns ~7.88e16 emissions; `assets(WELL)` returns 0.

Each field is read independently via `Promise.allSettled` and every rejection is reported via `environment.onError({ source: \"staking-fallback\", ... })`. Fallback triggers on a rejected `views.getStakingInfo()` OR a fully-zeroed views response (suspicious — a real deployment always has `cooldown` and `unstakeWindow` configured > 0). A new chain with zero stakers but real schedule constants still uses the views path.

### Moonbeam user-staking fallback — `src/actions/governance/getUserStakingInfo.ts`
`views.getUserStakingInfo(address)` also reverts on Moonbeam. Two narrow stkWELL reads cover the gap:
- `readUserStakingFromStkWell` — `stakersCooldowns(user)`, `getTotalRewardsBalance(user)`, `balanceOf(user)`
- `readScheduleFromStkWell` — `COOLDOWN_SECONDS()`, `UNSTAKE_WINDOW()`

Each rejection is surfaced via `environment.onError`. The function only returns `undefined` (drops the row) when every read in the fallback fails; partial failures yield 0n for that one field with a logged error.

## Test plan

- [ ] Once published, bump `@moonwell-fi/moonwell-sdk` in `moonwell-frontend` and verify on the Cloudflare preview build:
  - [ ] `/stake/base` Price tile reads ~\$0.0039 (not \$0.0522)
  - [ ] `/stake/optimism` Price tile matches Base
  - [ ] `/stake/moonbeam` page renders (no permanent skeleton) with consistent price + total staked
  - [ ] A Moonbeam user with a real stkWELL balance sees their stake / cooldown / pending rewards
  - [ ] Markets / Morpho rewards USD values still look correct
  - [ ] Moonriver staking page still shows MFAM-denominated price (not WELL)

## Notes for consumers

Every WELL-priced read now requires Base RPC access. If you create a `MoonwellClient` with only Moonbeam or Optimism configured, the SDK will fall back to its default public Base RPC. Configure Base explicitly in `networks` to honor your own RPC URL and `onError` handler.


lyoungblood:
## Summary

This PR centralizes the WELL/USD price lookup by introducing `getWellPriceFromBase()` (reads Base oracle's `getUnderlyingPrice(mWELL)`) and replacing 6 call sites that previously called `homeViewsContract.read.getGovernanceTokenPrice()`. It also restructures `getStakingInfo.ts` and `getUserStakingInfo.ts`, removing the per-environment home-environment lookup, and adds a `stkWELL` direct-read fallback in `getStakingInfo` when Moonbeam's views contract reverts.

## Issues

### 🔴 Critical

**1. Moonriver stkMFAM/MFAM rewards mispriced as WELL across all action paths.**
Moonriver has `governanceToken: "MFAM"` and `stakingToken: "stkMFAM"` (`src/environments/definitions/moonriver/contracts.ts:7`), with its own MFAM/USD oracle. Previously `homeViewsContract.read.getGovernanceTokenPrice()` resolved to Moonriver itself (no env's `chainIds` includes 1285) and returned MFAM/USD. The PR replaces every call with `getWellPriceFromBase()` which always returns Base WELL/USD. The wrong price now propagates through:
- `src/actions/governance/getStakingInfo.ts:89` → `:168` → stkMFAM `tokenPrice`/`totalSupplyUSD` = WELL/USD × stkMFAM totalSupply
- `src/actions/governance/getUserStakingInfo.ts:36` → `:110` → MFAM `tokenPrice` returned as WELL/USD
- `src/actions/core/markets/common.ts:58` → MFAM reward APR computed with WELL price (Moonriver is force-on-chain at line 24)
- `src/actions/core/user-rewards/common.ts:31` → MFAM `supplyRewardsUsd` / `borrowRewardsUsd` inflated by WELL:MFAM ratio

Suggested prompt:
\`\`\`text
In every place that calls getWellPriceFromBase() (markets/common.ts:58 and :510, core/user-rewards/common.ts:31, governance/getStakingInfo.ts:89, governance/getUserStakingInfo.ts:36, morpho/vaults/common.ts:325 and :1060, morpho/user-rewards/common.ts:336), do not unconditionally substitute the Base WELL price. Only use Base WELL when the env's environment.custom.governance.token === "WELL". For Moonriver (token === "MFAM"), keep the old behavior of reading the per-env homeViewsContract.read.getGovernanceTokenPrice(). Add a test that getStakingInfo and getMarkets on Moonriver still return a non-WELL price for the MFAM and stkMFAM tokens.
\`\`\`

**2. \`getUserStakingInfo\` lacks the stkWELL fallback that was added to \`getStakingInfo\`.**
The PR's stated fix is "restore staking info on Moonbeam by falling back to direct stkWELL reads when the views' getStakingInfo() reverts." But \`views.getStakingInfo()\` is *also* called at `src/actions/governance/getUserStakingInfo.ts:45` and gates the entire row at `:102` (\`if (!stakingInfoData) return [];\`). On Moonbeam the call still reverts, the user's \`getUserStakingInfo\`/\`balanceOf\` results are silently discarded, and Moonbeam users continue to see no user-staking data after this PR.

Suggested prompt:
\`\`\`text
In src/actions/governance/getUserStakingInfo.ts, when settled[2] (views.getStakingInfo()) is rejected on Moonbeam, fall back to a direct stkWELL read for the {cooldown, unstakeWindow} fields — the same fallback pattern used by getStakingInfoFromStkWell in getStakingInfo.ts. Reading stakingToken.read.COOLDOWN_SECONDS() and stakingToken.read.UNSTAKE_WINDOW() should be sufficient since that's all the downstream code at line 104 uses. Add a test that getUserStakingInfo on Moonbeam returns a non-empty result for a staked user.
\`\`\`

**3. stkWELL fallback queries \`assets()\` with the wrong key on Moonbeam.**
At `src/actions/governance/getStakingInfo.ts:59` the fallback calls \`stakingToken.read.assets([governanceToken.address])\` — passing WELL's address (\`0x511aB...\` on Moonbeam). Aave-style staking contracts (which Moonwell's stkWELL is a known fork of) initialize the \`assets\` mapping keyed by \`address(this)\` — i.e. the stkWELL contract itself. Looking at the ABI's \`configureAssets(uint128[], address[] underlyingAsset)\` signature combined with the Aave standard, the configured key is almost certainly the stkWELL address, not WELL. Wrong key → \`assets(WELL)\` returns \`(0,0,0)\` → \`emissionPerSecond = 0\` → APR collapses to 0 even though emissions are live. The "restore staking info on Moonbeam" fix produces a valid-shaped struct with wrong numbers, defeating its own purpose. **Recommend on-chain verification before shipping.**

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:59 the call stakingToken.read.assets([governanceToken.address]) passes the WELL address but Aave-fork staking contracts key the assets mapping by the stkWELL contract address itself (address(this) in Aave's StakedAave._initialize). Before merging, run a one-shot script that calls the deployed stkWELL on Moonbeam (0x8568A675384d761f36eC269D695d6Ce4423cfaB1) with .assets(WELL=0x511aB53F...) and .assets(stkWELL=0x8568A675...) and compare which one returns a non-zero emissionPerSecond. If the stkWELL key is the correct one, change the argument to stakingToken.address (the stakingToken token config address) and add a regression test asserting the Moonbeam APR returned by getStakingInfo is > 0.
\`\`\`

### 🟡 Major

**4. Fallback only triggers on rejection, not on stale/zero data.**
The changeset explicitly says "Moonbeam views contract returns stale data." If that staleness is a fulfilled-but-zero struct rather than a revert, `src/actions/governance/getStakingInfo.ts:113` (\`viewsStaking ?? (await getStakingInfoFromStkWell(environment))\`) skips the fallback because the value is truthy. Downstream the APR calculation at line 177 becomes \`((0 + 0) / 0 - 1) * 100 = NaN\`.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:113, the fallback should also engage when viewsStaking is structurally present but zeroed (specifically when totalSupply === 0n or emissionPerSecond === 0n). Replace the nullish-coalesce with an explicit check like: const useFallback = !viewsStaking || viewsStaking.totalSupply === 0n; const stakingInfo = useFallback ? await getStakingInfoFromStkWell(environment) : viewsStaking;
\`\`\`

**5. \`getStakingInfoFromStkWell\` drops 4 successful reads on any single failure.**
The single \`try { Promise.all([5 reads]) } catch { return undefined }\` at `src/actions/governance/getStakingInfo.ts:52-73` means a transient RPC error on any one read (e.g. throttled \`assets()\` call) erases the entire fallback, returning the chain to the broken state the PR is meant to fix. Worse, there's no \`console.warn\` or \`environment.onError(...)\` — diagnosing a Moonbeam staking dropout requires manual debugging.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfoFromStkWell (around line 52), change the single try/catch with Promise.all to Promise.allSettled and read each field independently, defaulting to 0n when a specific call rejects. Also log a console.warn or environment.onError({source: 'staking-fallback', chainId: environment.chainId}, error) so silent failures are observable.
\`\`\`

**6. \`Promise.all\` in morpho paths can fail entirely when Base oracle hiccups.**
At `src/actions/morpho/vaults/common.ts:325`, `:1060`, and `src/actions/morpho/user-rewards/common.ts:336`, \`getWellPriceFromBase()\` is wrapped in a bare \`Promise.all\`. Before the PR, the analogous call hit the *home* chain (Moonbeam for Optimism, etc.), so Base RPC issues didn't affect non-Base flows. After the PR, every chain's morpho vault/reward query couples its availability to Base's oracle/RPC. The first call site at \`:1060\` is the worst because its enclosing for-loop has no try/catch — a Base outage rejects vault data for all queried chains, not just Base.

Suggested prompt:
\`\`\`text
In src/actions/morpho/vaults/common.ts:325, src/actions/morpho/vaults/common.ts:1060, and src/actions/morpho/user-rewards/common.ts:336, wrap the getWellPriceFromBase() call in a .catch(() => 0n) before passing it to Promise.all, matching the pattern used by getStakingInfo.ts:89. This restores per-call failure isolation so a Base oracle blip doesn't cascade into non-Base flows.
\`\`\`

**7. Centralized \`.catch(() => 0n)\` silently zeroes USD output for ALL chains on a single Base failure.**
At `src/actions/governance/getStakingInfo.ts:89` (and analogously at `getUserStakingInfo.ts:36`) a Base RPC blip becomes \`governanceTokenPrice = 0n\`, and *every* returned \`StakingInfo\` reports \`tokenPrice: 0\` and \`totalSupplyUSD: 0\` — including chains whose own RPC succeeded. The UI cannot distinguish "price unavailable" from "actually \$0." Previously each chain fetched its own price; failures were chain-local.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:89 and src/actions/governance/getUserStakingInfo.ts:36, when the .catch fires, also call client.environments.base?.onError?.(error, {source: 'well-price', chainId: base.id}) so consumers can detect that USD valuations should be treated as missing rather than zero.
\`\`\`

**8. \`getWellPriceFromBase()\` ignores user-configured Base RPC at 6 call sites.**
The function accepts \`client?: MoonwellClient\` but the 6 non-staking call sites (markets, user-rewards, morpho/vaults, morpho/user-rewards) only have \`environment\` in scope and pass nothing. The function then defaults to \`publicEnvironments.base\` (`src/actions/governance/getWellPrice.ts:22`), bypassing any custom Base RPC URL or \`onError\` configured in the user's MoonwellClient. Also, the line uses a \`(client?.environments as { base?: Environment } | undefined)?.base\` cast that assumes the user keyed Base specifically as \`"base"\` — fork test setups using a different alias silently hit mainnet.

Suggested prompt:
\`\`\`text
Refactor src/actions/governance/getWellPrice.ts to accept a baseEnvironment: Environment | undefined parameter instead of a client. Then at each call site (markets/common.ts:58 etc.) compute the correct Base env from the surrounding scope (e.g. resolve via the function's environments parameter or find an env where chainId === base.id) and pass it explicitly. Drop the `as { base?: Environment }` type assertion. This honors user-configured RPCs and fork environments uniformly across all call sites.
\`\`\`

**9. Falsy-zero guard treats legitimate \`0n\` returns as failure.**
`src/actions/core/user-rewards/common.ts:48` and `src/actions/core/markets/common.ts:513` check \`!governanceTokenPriceRaw\`. Since BigInt \`0n\` is falsy in JS, when \`getWellPriceFromBase()\` falls back to its \`0n\` return (missing config) or the oracle legitimately returns 0, these branches treat it as failure: user-rewards returns \`[]\`, markets throws. Combined with the silent-0 path in finding 7, the same condition produces inconsistent symptoms across actions.

Suggested prompt:
\`\`\`text
In src/actions/core/user-rewards/common.ts:48 and src/actions/core/markets/common.ts:513, change the falsy checks to explicit undefined checks (e.g. governanceTokenPriceRaw === undefined) so a legitimate 0n price doesn't trigger the failure path. Same pattern review across all sites using || to coerce away BigInt zero.
\`\`\`

### 🔵 Minor

**10. \`getStakingInfoFromStkWell\` returns undefined on missing \`governanceToken\` config but never logs it.**
`src/actions/governance/getStakingInfo.ts:50` silently returns undefined if \`governanceToken\` lookup fails — same observability gap as finding 5.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:50, add an environment.onError or console.warn before return undefined; so misconfigured environments are diagnosable.
\`\`\`

**11. Type assertions violate project's "No type assertions" rule.**
The project CLAUDE.md says "No type assertions (\`as Type\`) — restructure to avoid casts," but the PR adds:
- \`src/actions/governance/getWellPrice.ts:22\` — \`as { base?: Environment } | undefined\`
- \`src/actions/governance/getWellPrice.ts:27\` — \`as bigint\`
- \`src/actions/governance/getStakingInfo.ts:43-45\` — \`as keyof TokensType<typeof environment> | undefined\`
- \`src/actions/governance/getStakingInfo.ts:62\` — \`as readonly [bigint, bigint, bigint]\`
- \`src/actions/governance/getStakingInfo.ts:147-149\` — \`as StakingInfoStruct | undefined\`

Suggested prompt:
\`\`\`text
Remove the as type assertions added in src/actions/governance/getWellPrice.ts and src/actions/governance/getStakingInfo.ts. Where viem's inferred type is correct, just remove the cast. Where the cast was needed to narrow a return value, restructure the type at the source (e.g. infer the StakingInfoStruct from the viem-generated return type of getStakingInfo() instead of declaring a separate local interface).
\`\`\`

**12. Non-null assertions (\`!\`) violate project conventions.**
At `src/actions/governance/getStakingInfo.ts:141`, `:145`, `:147`, `:150` the PR still uses non-null assertions on \`envStakingInfo[index]![0]\` etc., which is explicitly banned in CLAUDE.md: "No non-null assertions (\`!\`) — use optional chaining or nullish coalescing."

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts replace each envStakingInfo[index]![0] / envStakingInfo[index]![1] with optional-chaining + fallback, e.g. envStakingInfo[index]?.[0] ?? undefined. The downstream if (!envStakingInfoData ...) guard already handles undefined.
\`\`\`

**13. \`assetData as readonly [bigint, bigint, bigint]\` assumes viem tuple semantics.**
viem returns multi-output reads as tuples currently, but the ABI's \`assets\` outputs are fully-named (\`emissionPerSecond\`, \`lastUpdateTimestamp\`, \`index\`), which in some viem versions becomes an object. The cast silently masks that mismatch if it ever changes.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:62 remove the as readonly [bigint, bigint, bigint] and rely on viem's inferred return type. If destructuring is awkward because of object vs tuple semantics, use assetData[0] (tuple) or assetData.emissionPerSecond (object) — whichever the inferred type yields.
\`\`\`

**14. New cross-chain dependency for Moonbeam-only or Optimism-only clients is undocumented.**
A consumer instantiating \`createMoonwellClient\` with only Moonbeam configured now silently makes Base-mainnet RPC calls via \`publicEnvironments.base\`. This is a privacy and reliability change not mentioned in the changeset.

Suggested prompt:
\`\`\`text
Add a note to .changeset/well-price-from-base-oracle.md explaining that all WELL-priced reads now require Base RPC access, and that consumers should configure Base in their MoonwellClient (or accept the SDK's default public Base RPC).
\`\`\`

**15. Asymmetric resilience: live data has fallback, x28-historical read does not.**
On Base, `src/actions/governance/getStakingInfo.ts:155` still drops the entire Base row if the archive-block read at 34149943 rejects. The PR improves resilience for the current-block read (stkWELL fallback) but leaves the historical read brittle — a non-archive RPC kills Base output despite live data succeeding.

Suggested prompt:
\`\`\`text
In src/actions/governance/getStakingInfo.ts:155, if the archive-block getStakingInfo() at block 34149943 rejects, consider proceeding without it rather than dropping the whole Base row. The x28-proposal historical totalSupply is used for what — verify it's actually load-bearing for correctness, and if not, treat its absence as 0n rather than as a hard gate.
\`\`\`

## Positive observations

- The high-level intent — replacing per-chain governance price reads with a single authoritative Base oracle source — is the right architectural call given the changeset's description of broken upstream views contracts.
- The new tuple shape in \`getStakingInfo\` (\`[stakingInfo, historicalStaking]\`) is cleaner than the previous variable-length array and removes a subtle index-misalignment risk after the \`.filter(s => s.status === "fulfilled")\` deletion.
- \`getStakingInfoFromStkWell\` is correctly scoped as a fallback (only triggers when views fails) rather than an unconditional path, preserving the working flow on Base.
- The \`.catch(() => 0n)\` on \`getWellPriceFromBase\` in staking flows is defensive — though the centralization downside is noted in finding 7.

## Suggestions

- The new \`getWellPriceFromBase()\` is global state in disguise. If you keep it, document the call-site contract (Promise.all vs allSettled, client passing) and ideally enforce it via the signature (e.g., require an \`Environment\` argument).
- Consider extracting the home-environment lookup to a shared utility so the files that still duplicate it (\`markets/common.ts:41\`, \`:503\`, \`core/user-rewards/common.ts:19\`, \`morpho/user-rewards/common.ts:324\`, \`morpho/vaults/common.ts:312\`, \`:1048\`) stay consistent.

Suggested prompt:
\`\`\`text
Audit the codebase for the duplicated home-environment lookup pattern (Object.values(publicEnvironments) as Environment[]).find((e) => e.custom?.governance?.chainIds?.includes(environment.chainId)) || environment (appears at markets/common.ts:41/503, core/user-rewards/common.ts:19, morpho/user-rewards/common.ts:324, morpho/vaults/common.ts:312/1048). Extract to a single utility in src/environments/utils/ so its semantics are consistent across all callers.
\`\`\`

lyoungblood:
Pushed `d550809` addressing review feedback. Summary of what changed:

## Critical (all addressed)

**#1 — Moonriver MFAM mispricing.** Replaced `getWellPriceFromBase()` with a dispatcher `getGovernanceTokenPriceFor(env, baseEnv?)` that branches on `env.custom?.governance?.token`. WELL → Base oracle; MFAM (Moonriver) → its own per-chain `views.getGovernanceTokenPrice()` (which is the home for MFAM). Wired through all 7 call sites — each now passes the local `environment` so the dispatch happens per-chain.

**#2 — getUserStakingInfo lacks fallback.** Added a stkWELL-direct fallback for user staking too. Both `views.getUserStakingInfo()` and `views.getStakingInfo()` revert on Moonbeam (verified via eth_call), so I added two narrow helpers — `readUserStakingFromStkWell` (uses `stakersCooldowns`, `getTotalRewardsBalance`, `balanceOf`) and `readScheduleFromStkWell` (`COOLDOWN_SECONDS`, `UNSTAKE_WINDOW`). Either views-fulfilled-struct or the fallback satisfies the downstream code now.

**#3 — `assets()` wrong key.** Verified on-chain — the reviewer was right:
```
assets(WELL=0x511a…)   → emissionPerSecond = 0
assets(stkWELL=0x8568…) → emissionPerSecond = 7.88e16 (~2.49M WELL/year)
```
Switched the call to pass `stakingTokenAddress` (= `stkWELL`). Added a doc comment recording the Aave-fork convention (`address(this)` in `StakedAave._initialize`).

## Major (all addressed)

- **#4** — Fallback now also engages when views returns a fulfilled-but-zeroed struct (`viewsStaking.totalSupply === 0n`), not only on rejection.
- **#5** — `getStakingInfoFromStkWell` reads each field with `Promise.allSettled` and only bails out when `totalSupply` itself fails (totalSupply is the load-bearing field). First rejection is reported via `environment.onError(..., { source: "staking-fallback" })`.
- **#6** — All 3 morpho `Promise.all` sites now `.catch(() => 0n)` on the Base price so a Base RPC blip doesn't cascade into non-Base flows.
- **#7** — Price-fetch rejections in `getStakingInfo` / `getUserStakingInfo` are surfaced via `environment.onError(..., { source: "governance-token-price" })`; the `0n` fallback is still applied so other chains' rows aren't dropped.
- **#8** — Helper now accepts `(environment, baseEnvironment?)`. Staking sites read `client.environments.base` once and pass it through. Non-staking sites just pass `environment`. No more global `client` capture.
- **#9** — Replaced `!governanceTokenPriceRaw` / `|| 0n` with `=== undefined` / `?? 0n` in `core/markets/common.ts`, `core/user-rewards/common.ts`, and both morpho sites so a legitimate `0n` doesn't trigger failure paths.

## Minor

- **#10** — `onError` call added when stkWELL config is missing in the fallback (same observability pattern as #5/#7).
- **#11/#12** — Removed the type assertions and non-null assertions called out. Replaced `as StakingInfoStruct` with `isStakingInfoStruct()` runtime guards. A handful of `as Record<string, ...>` casts on dynamically-keyed `config.tokens` lookups remain — the SDK's generic types don't expose specific token keys, and the alternative (extending the generics) was out of scope for this PR.
- **#13** — Removed the readonly tuple cast on `assets` return; the inline comment documents that viem returns multi-output reads as a tuple even when the ABI names the outputs.
- **#14** — Changeset updated with a note that all WELL-priced reads now require Base RPC access (default public Base RPC if not configured).

## Not changed (with rationale)

- **#15 (asymmetric resilience on the x28 historical block)** — Left as-is. The historical-block `getStakingInfo` at block 34149943 looks load-bearing for Base's pre/post-proposal accounting; degrading it to `0n` could silently corrupt downstream math. Would prefer a follow-up where someone owning that math confirms what falls back safely.
- **The home-environment lookup duplication** in the Suggestions section is real but a refactor for its own PR.

## Dev chore

`package.json`: flipped `pnpm test` to `vitest run` (one-shot) and added `pnpm test:watch` for the prior watch-mode behavior. The previous `dev` subcommand never exited, which hangs CI and agent harnesses.

---

Re-run the test suite when convenient — note that several tests on `main` are already failing (per discussion), so this PR's CI test failure isn't introduced by these changes.

lyoungblood:
Pushed `a098a73` addressing a second pass of code-review findings (recall-mode review).

**Observability gaps** — every stkWELL-fallback helper now fires `environment.onError` for every rejected read, not just the first or only when *all* reads fail. A single throttled `balanceOf()` or `COOLDOWN_SECONDS()` call no longer silently zeroes a field with no telemetry. Same for `getWellPriceFromBaseOracle` when the Base env config is missing `MOONWELL_WELL` or `oracle`, and for the three `morpho` + one `markets` `.catch(() => 0n)` sites — each now logs via `environment.onError` before returning 0n, so a Base oracle outage materializes as an observable error instead of silent \$0 USD values.

**Correctness** — `getStakingInfoFromStkWell` no longer treats a *fulfilled* `totalSupply === 0n` as failure. Previously, a brand-new chain with zero stakers would have been dropped from the result list. The trigger for the views→fallback switch is also tightened: fallback fires only on a *fully-zeroed* views response (`cooldown=0 AND unstakeWindow=0 AND totalSupply=0 AND emission=0`), not just `totalSupply=0n`. Empty new chains with real schedule constants stay on the views path.

**Cleanup** — removed the dead `if (!entry) return []` guards in both flatMap loops (`envStakingInfo` is built via `Promise.all`/`Promise.allSettled` over the same `envsWithStaking` array, so index access is always defined). Dropped the dead `governanceTokenPriceRaw ?? 0n` after the catch (the catch already guarantees `bigint`).

PR description has been refreshed too — it was still referencing the old `getWellPriceFromBase` name and didn't mention Moonriver/MFAM dispatch, the user-staking fallback, or the `assets()` key fix. The Test plan now also includes Moonriver as a regression check.

PR #280, opened 2026-05-26
add Ethereum Mainnet staking

bprofiro:
### Summary                                                                                                                                                                                                                                        
                                                                                            
  - Wire Ethereum mainnet stkWELL + the new staking views contract (0xF5f2ae75…Ce5401) so getStakingInfo, getUserStakingInfo, and getStakingSnapshots return Ethereum entries alongside Base / Optimism / Moonbeam.                              
  - Skip Ethereum in getUserVotingPowers / getDelegates with a one-time warning — the Eth views contract is staking-only and doesn't expose getUserVotingPower.                                                                                  
  - Make getUserBalances fall through to per-token balanceOf reads if getTokensBalances reverts (the staking-only Eth views contract doesn't implement it).                                                                                      
                                                                                                                                                                                                                                                 
## Test plan                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                 
  - pnpm typecheck                                                                                                                                                                                                                               
  - pnpm test — getUserVotingPowers suite skips mainnet and passes on other chains
  - Manual: call getStakingInfo / getUserStakingInfo against an Eth RPC and confirm an Ethereum entry comes back                                                                                                                                 
  - Manual: call getUserBalances on Ethereum and confirm balances populate via the fallback path


apokusin (dismissed):
looks tentatively good to me, just waiting for claude code-review to come back

apokusin:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

apokusin:
### Follow-ups (non-blocking)

These were surfaced during review but are below the bar to block the PR. Logging them as optional follow-ups for awareness.

1. `getUserBalances` issues a guaranteed-to-revert `getTokensBalances` on Ethereum's staking-only views contract every call, then catches and falls back. Functionally correct, but the branch commit titled "dont call gettokenbalances" suggests the intent may have been to skip the call. Consider a capability/chain guard (as `getUserVotingPowers.ts` does with `STAKING_ONLY_VIEWS_CHAIN_IDS`) to avoid the wasted RPC + `onError` on every Ethereum balance fetch.

https://github.com/moonwell-fi/moonwell-sdk/blob/f38eb7c0bb6cfa14234cf261f67773c8f3672df7/src/actions/core/getUserBalances.ts#L82-L103

2. Comment frames the limitation as ABI-level ("exposes getUserStakingVotingPower but not ... getUserVotingPower"), but `coreViewsAbi.ts` is a single shared ABI defining both functions — this is the deployed contract's runtime revert behavior, not an ABI difference. Minor reword for accuracy.

https://github.com/moonwell-fi/moonwell-sdk/blob/f38eb7c0bb6cfa14234cf261f67773c8f3672df7/src/actions/governance/getUserVotingPowers.ts#L15-L18

3. `getDelegates` hardcodes `targetChainIds` excluding mainnet, which diverges from `WELL.chainIds` (now includes mainnet). The result is correct (Eth views is staking-only), but the same exclusion now lives in two places; this was previously raised on #270. Consider deriving from a capability check.

https://github.com/moonwell-fi/moonwell-sdk/blob/f38eb7c0bb6cfa14234cf261f67773c8f3672df7/src/actions/governance/getDelegates.ts#L31-L38

4. Test comment claims it gates "the same way getStakingSnapshots.test.ts does", but that file additionally requires `lunarIndexerUrl`. The narrower gate here is correct for `getStakingInfo`; just a misleading comment.

https://github.com/moonwell-fi/moonwell-sdk/blob/f38eb7c0bb6cfa14234cf261f67773c8f3672df7/src/actions/governance/getStakingInfo.test.ts#L14-L17

🤖 Generated with [Claude Code](https://claude.ai/code)

avp1598 (approved):

PR #288, opened 2026-05-27
Recognize Ethereum-home multichain proposals and enable Moonbeam-wall…

bprofiro:
…et voting on them


lyoungblood:
# Staff Engineer Review

> _Recognize Ethereum-home multichain proposals and enable Moonbeam-wallet voting on them_

## Summary

Three independently-motivated governance changes bundled into one PR:

1. **`isMultichainProposal`** now recognizes both the Moonbeam (`0xc8e2…29f3`) and Ethereum (`0x98f3…288b`) Wormhole Core Bridges, so proposals authored on the Ethereum multigov hub stop being misclassified as single-chain.
2. **`getProposal`** routes `chainId: 1` (Ethereum) calls through `getMoonbeamProposal` against the Moonbeam env, since the lunar indexer serves both chains. Previously, deep-linking to a chainId=1 proposal bailed out at the `!moonbeam && !moonriver` guard and returned `undefined`.
3. **Moonbeam voting-source policy change**: raw WELL (`tokenVotes`) is masked to zero on Moonbeam in both `getUserVotingPowers` and `getDelegates`. The Moonbeam contracts config gains the new satellite `voteCollector` (`0xB8A7…40E5`).

No public API surface is removed or renamed — additions only — but `getUserVotingPowers` / `getDelegates` *return values change semantically* on Moonbeam (token-prefixed fields read as zero), which the `minor` changeset correctly captures.

Quality gate is green; coverage +0.69pp lines. Typecheck clean. Unit tests pass locally; the 8 integration failures in my sandbox were RPC-allowlist artifacts, not regressions.

## Issues

### 🟡 Major

**1. Zero test coverage for the new `getProposal` mainnet routing — `src/actions/governance/proposals/getProposal.ts:59-70`.**

The new branch is the headline feature of this PR yet has no unit test. The existing `getProposal.test.ts` suite never exercises `network: "mainnet"` or `chainId: 1`. A single fetch mock + a "moonbeam env missing → undefined" guard test would close the regression risk. Without these, a future refactor of `getEnvironmentFromArgs` or the chainId routing could silently re-break Ethereum deep-links.

````text
Add tests to src/actions/governance/proposals/getProposal.test.ts covering the new mainnet routing branch in getProposal.ts:59-70:

1. A test that calls getProposal with `chainId: 1` against a client containing both an `ethereum`/`mainnet` env and a `moonbeam` env, mocks fetchProposal to return a chainId=1 ApiProposal, and asserts fetchProposal was called with the moonbeamEnv and chainId=1 — not with the ethereum env.

2. A test that calls getProposal with `network: "mainnet"` (no explicit chainId) and asserts the routing still hits chainId=1 only (does NOT fall back to 1284) because the caller picked Ethereum explicitly.

3. A test that calls getProposal with chainId=1 on a client whose environments object has no `moonbeam` key and asserts it returns undefined without invoking fetchProposal.

Reuse the existing moonbeamEnv/moonriverEnv/baseApiProposal fixtures at the top of the file. Add the new tests inside a new `describe("getProposal mainnet routing", ...)` block.
````

**2. Key-name coupling in the mainnet branch — `src/actions/governance/proposals/getProposal.ts:60-61`.**

```ts
const moonbeamEnv = (client.environments as { moonbeam?: Environment })
  .moonbeam;
```

This relies on `"moonbeam"` being the literal key in `client.environments`. The whole governance subsystem otherwise looks up by chainId (`Object.values(...).find(e => e.chainId === moonbeam.id)`, e.g. `common.ts:303`, `common.ts:610`, `getEnvironmentFromArgs` itself), or uses `publicEnvironments.moonbeam` directly when the canonical env is needed (`getDelegates.ts:26`). A consumer-built client that re-keys envs would break here in a way the type system can't catch.

````text
In src/actions/governance/proposals/getProposal.ts:59-70, replace the literal-key access `(client.environments as { moonbeam?: Environment }).moonbeam` with a chainId-based lookup so the routing doesn't assume a specific env key name. Match the pattern used in common.ts:303 / common.ts:610:

const moonbeamEnv = Object.values(
  client.environments as Record<string, Environment>,
).find((e) => e.chainId === moonbeam.id);

Keep the `if (!moonbeamEnv) return undefined;` guard. `moonbeam` is already imported from viem/chains.
````

**3. The `voteCollector` wire-up has no test pinning the behavior change it unlocks — `src/environments/definitions/moonbeam/contracts.ts:17`.**

Adding `voteCollector` to Moonbeam contracts is operative: it changes the `xcEnvironments` filter in `getProposalsOnChainData` (`common.ts:434-445`) and `getCrossChainProposalData` (`common.ts:609-620`), since both require `voteCollector` to include an env in the cross-chain vote read. Previously Moonbeam was excluded; now an Ethereum-home (chainId=1) proposal's `votesCollected` will read against the Moonbeam vote collector too. That's almost certainly the intent of the PR title, but nothing locks it down — a test that doesn't fail is one that doesn't really exist. A focused unit test (or an integration assertion that the Moonbeam env passes the filter) would prevent a future removal/typo from silently turning off Moonbeam vote collection.

````text
Add a unit test (in src/actions/governance/proposals/common.test.ts or a new common.votesCollected.test.ts) that mocks a Moonbeam-shaped governanceEnvironment with `multichainGovernor` and a stub `chainVoteCollectorVotes`, plus a minimal mock `publicEnvironments` exposing both `moonbeam` (now with voteCollector) and `base`. Call getProposalsOnChainData with a multichain proposal (targets containing WORMHOLE_CONTRACT) and assert that the Moonbeam vote collector is exercised — i.e. chainVoteCollectorVotes is called with the Moonbeam wormhole chainId. This pins down that adding `voteCollector` to Moonbeam contracts.ts:17 actually flows through the filter at common.ts:434-445.
````

### 🔵 Minor

**4. Redundant `.toLowerCase()` on every call — `src/actions/governance/proposals/common.ts:106-112`.**

Both entries in `MULTICHAIN_WORMHOLE_BRIDGES` are already lowercase string literals, so the `bridge.toLowerCase()` on line 110 is dead work. Also, the new implementation lower-cases *every* target up front instead of short-circuiting on the first match like the old `.some()` did. Low impact (target lists are short), but the symmetric version reads cleaner:

````text
In src/actions/governance/proposals/common.ts:106-112, simplify isMultichainProposal so it (a) doesn't lower-case the bridge list redundantly (the constant values are already lowercase) and (b) preserves the short-circuit semantics of `.some()` rather than fully mapping `targets` up front. Switch MULTICHAIN_WORMHOLE_BRIDGES to a `ReadonlySet<string>` derived once at module load, then:

export const isMultichainProposal = (targets?: string[]): boolean =>
  targets?.some((t) => MULTICHAIN_WORMHOLE_BRIDGES.has(t.toLowerCase())) ?? false;
````

**5. `getMoonbeamProposal` is now a misnomer — `src/actions/governance/proposals/getProposal.ts:93`.**

After this PR it handles both Moonbeam-historical and Ethereum-multigov proposals via the Moonbeam env. The JSDoc explains this clearly, but the name lies. Worth a follow-up rename to `getMultigovProposal` once the surface is small — every additional caller cements the wrong name.

**6. Policy mask duplicated across two files — `src/actions/governance/getUserVotingPowers.ts:111-114` and `src/actions/governance/getDelegates.ts:67-71`.**

Both places hard-code `env.chainId === moonbeam.id` as the "mask raw WELL" predicate. If the policy expands (e.g. mask on more chains), or reverses, two sites drift. A single exported constant — e.g. `RAW_WELL_MASKED_CHAINS: ReadonlySet<number>` — would keep both call sites honest.

````text
Extract the "Moonbeam raw-WELL is not an eligible voting source" policy into a single source of truth. Create a new constant in src/actions/governance (or extend a suitable existing module) that exports:

export const RAW_WELL_MASKED_CHAINS: ReadonlySet<number> = new Set([moonbeam.id]);

Then have getUserVotingPowers.ts:111-114 and getDelegates.ts:67-71 both consult it via `RAW_WELL_MASKED_CHAINS.has(env.chainId)` instead of `env.chainId === moonbeam.id`. Keep the existing tests green; no new tests required.
````

## Positive observations

- Two **separate** changesets for two logically independent changes — exactly the right granularity. Captures the consumer-facing semantic change without forcing a major bump.
- New test **"matches when Wormhole bridge is buried in a long targets list"** is a thoughtful guard against the naive "only check `targets[0]`" failure mode.
- Counterpart sanity test **"non-Moonbeam chains keep raw WELL tokenVotes in voting power"** is exactly the right shape — preventing a future regression where someone "generalizes" the mask and accidentally globalizes it.
- `ZERO_TOKEN_VOTES` is module-scoped `as const` and reused by reference, rather than allocated per-row in the `.map`.
- Inline comments explain the *why* (policy change, indexer dual-fan-out) rather than the *what* — matches the house style.
- Quality gate moved forward, not held. Coverage and functions both up, no audits regressed.

## Suggestions (non-blocking)

- Once a second chain joins the "raw token masked" list (or once Moonbeam graduates back), the constant extraction from #6 will pay for itself. Worth doing now even speculatively because the policy is exactly one chain wide today — i.e., easiest possible diff.
- After the rename in #5, consider co-locating `getMoonbeamProposal` / `getMoonriverProposal` into named module exports under `proposals/` so the routing in `getProposal` becomes a single switch on the proposal's home chain rather than a layered if-ladder. Out of scope for this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

**Comment by apokusin** [comment]
## Follow-up review on `12465ac` ("address comments")

The follow-up addresses 5 of 6 prior findings (Staff #1, #2, #3, #4, #6). The remaining Staff #5 (`getMoonbeamProposal` misnomer) was already marked as follow-up. Typecheck is clean.

**Recommendation: ship as-is.**

Two minor observations on the new code, neither blocking:

### 🔵 Minor — residual cast in the fix for #2 (`src/actions/governance/proposals/getProposal.ts:60-62`)

```ts
Object.values(client.environments as Record<string, Environment>)
```

The fix swaps `as { moonbeam?: Environment }` for `as Record<string, Environment>`. Strictly, CLAUDE.md's "no type assertions in `src/`" still applies. In practice this one is hard to avoid since `client.environments` is a generic type parameter, so it's a wash. Skip unless you want to be pedantic.

### 🔵 Minor — voteCollector test has no negative control (`src/actions/governance/proposals/common.test.ts:327-367`)

The new wire-up test asserts `chainVoteCollectorVotes` is called with the Moonbeam wormhole chainId, but doesn't prove that removing `voteCollector` from the mocked Moonbeam env would make it *not* be called. If the filter at `common.ts:434-445` were already broken in a way that lets every env through, the test would still pass.

A short counterpart — re-run with `publicEnvironments.moonbeam.contracts.voteCollector` unset and assert the Moonbeam wormhole chainId is **not** in `mock.calls` — would actually pin the wiring against future regressions.

```text
In src/actions/governance/proposals/common.test.ts, add a counterpart test inside the "getProposalsOnChainData voteCollector wire-up" describe block. Use vi.doMock or a second vi.mock factory to swap publicEnvironments.moonbeam.contracts to {} (no voteCollector). Re-run getProposalsOnChainData with the same env and proposal as the existing test, and assert that chainVoteCollectorVotes was NOT called with [MOONBEAM_WORMHOLE_CHAIN_ID, 7n] — only the Base call should appear. This pins the filter behavior in both directions.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

apokusin (approved):

PR #290, opened 2026-05-28
fix eth proposals timeline

bprofiro:
## Summary

Fix the Ethereum-hub proposal timeline. `getProposalsOnChainData` was reading on-chain state against the caller's `multichainGovernor` (Moonbeam) instead of the proposal's home governor (Ethereum for chainId=1), so `eta` stayed `0` and `votesCollected` never flipped — leaving the proposal-detail timeline stuck without a timelock countdown.

## Changes

- Resolve each proposal's home env from `publicEnvironments` and read `state` / `proposalData` (eta) from the right chain's `multichainGovernor`.
- Derive `votesCollected` from the governor's own state machine (`state > MultichainVoteCollection`) instead of AND-ing per-satellite `chainVoteCollectorVotes`. The old gate pinned `false` forever for satellites with zero turnout.
- API-derived state fallback when on-chain reads fail — terminal proposals no longer surface as `Pending(0)` on RPC outages.
- Skip synthetic eta when only `proposals()` fails — avoids pairing real on-chain state with a fabricated countdown.
- `Promise.allSettled` for parallel state/proposals reads with per-read error handling routed through `onError`.
- Add `wormhole.chainId: 24` to Optimism (was missing — would silently exclude OP from satellite enumeration).

Moonbeam-hub proposals: untouched (local-env code path preserved).

## Test plan

- [ ] `npm run test` clean
- [ ] `getProposal({ proposalId: 169, chainId: 1 })` returns the real on-chain eta and state
- [ ] Frontend shows full 5-step timeline + timelock countdown for proposal 169


apokusin:
## Code Review

### Summary

This PR resolves on-chain timeline data for Ethereum-hub multigov proposals fetched through the Moonbeam governance env: `getProposalsOnChainData` now resolves each proposal's home environment and reads `state` / `proposalData` (eta) / `chainVoteCollectorVotes` against that env's `multichainGovernor`, with per-read `try/catch` + `onError` wiring and an API-derived state fallback. The core approach is sound and the error-handling/telemetry improvements are genuine. **No public API signatures change**, so there's no consumer-facing breaking change — but the change to `src/environments/definitions/ethereum/custom.ts` (`governance.chainIds`) has effects outside the governance subsystem that I'd want resolved before merge.

> _Updated after re-tracing call paths: Issue #1 is redirected to the always-on `getUserRewards` path (the markets path I originally cited is fallback-only), and the two satellite-set issues are merged into one, with the Moonbeam half flagged as needing on-chain verification rather than asserted as a confirmed bug._

### Issues

#### 🔴 Critical

**1. Adding `moonbeam.id` to Ethereum's `governance.chainIds` re-points the Moonbeam env's `homeEnvironment` to Ethereum, mispricing Moonbeam native-token user rewards** — `src/environments/definitions/ethereum/custom.ts:8`

`custom.governance.chainIds` is **not** governance-private — it's used as a "home environment" membership predicate by five non-governance call sites:

- `src/actions/core/user-rewards/common.ts:21` — **always-on** (no Lunar fast-path)
- `src/actions/morpho/user-rewards/common.ts:158`
- `src/actions/core/markets/common.ts:43` and `:505` — on-chain **fallback path only** (guarded by `lunarIndexerUrl`)
- `src/actions/morpho/vaults/common.ts:316` and `:1054`

Each does `Object.values(publicEnvironments).find(e => e.custom?.governance?.chainIds?.includes(environment.chainId)) || environment`. `publicEnvironments` is ordered `base, moonbeam, moonriver, optimism, ethereum, …` (`src/environments/index.ts:177`). Before this PR nothing listed `1284`, so the **Moonbeam** env matched nothing and fell back to itself. Now Ethereum lists `1284`, so for the Moonbeam env `find` returns **Ethereum**.

The concrete, always-on impact is in **core user rewards**: `getUserRewards → getUserRewardsData` resolves `homeEnvironment` on every call and feeds `homeViewsContract.getNativeTokenPrice()` into `nativeTokenPrice`, which is then mapped to the zero-address (native gas token) reward of the *current* env (`user-rewards/common.ts:91-98, 127, 135`). So Moonbeam GLMR-denominated rewards are now valued using **Ethereum's** native price every call. Core markets is affected only on the on-chain fallback path (when `lunarIndexerUrl` is unset or the Lunar fetch throws). Base/Optimism and Morpho are unaffected — Moonbeam precedes Ethereum in the ordering, and there's no Moonbeam Morpho deployment.

```text
In src/environments/definitions/ethereum/custom.ts, stop overloading custom.governance.chainIds for the Ethereum-hub Wormhole satellite list. That field is also a home-environment membership predicate used by src/actions/core/user-rewards/common.ts:21 (always-on), src/actions/morpho/user-rewards/common.ts:158, src/actions/core/markets/common.ts:43 and 505, and src/actions/morpho/vaults/common.ts:316 and 1054. Adding moonbeam.id (1284) makes the Moonbeam environment resolve its homeEnvironment to Ethereum instead of itself, so getUserRewards values Moonbeam native (GLMR) rewards using Ethereum's getNativeTokenPrice. Introduce a separate field for the governance vote-collector satellites (or derive the satellite set inside getProposalsOnChainData from environments exposing both custom.wormhole.chainId and contracts.voteCollector) and revert ethereum custom.governance.chainIds to []. Add a regression test asserting the Moonbeam environment's homeEnvironment still resolves to Moonbeam.
```

#### 🟡 Major

**2. The Ethereum satellite set is hand-listed and not validated against the wired environments** — `src/environments/definitions/ethereum/custom.ts:8`, `src/actions/governance/proposals/common.ts:474-526`

`chainIds = [moonbeam, base, optimism]` drives the `votesCollected` enumeration, which filters to envs exposing both `custom.wormhole.chainId` and `contracts.voteCollector`, then ANDs `chainVoteCollectorVotes > 0` across the survivors (`.every(...)`, lines 523-526). The hand-maintained list is inconsistent with the actual wiring in two ways:

- **Optimism is listed, but `src/environments/definitions/optimism/custom.ts` has no `wormhole` block**, so the filter at lines 480-491 silently drops it. `votesCollected` then ANDs only Moonbeam + Base — contradicting the changeset's "non-zero votes on **every** satellite." If Optimism is a genuine collector, a proposal can report `votesCollected: true` while Optimism is still uncollected. _(Verified from source.)_
- **Moonbeam is included**, but the changeset itself calls Moonbeam the execution/temporal chain, not a vote source. If the Ethereum `multichainGovernor` has **no** collector registered for Wormhole chain 16, `chainVoteCollectorVotes([16, id])` returns zeros and the AND can never pass — leaving the timeline stuck on "Vote Collection." _I can't verify the on-chain collector registration from the SDK — this needs checking against the deployed contract._

Deriving the satellite set from environments that actually expose a `voteCollector` (the predicate already applied at lines 480-491) would remove the hand-maintained list and close both failure modes.

```text
In src/actions/governance/proposals/common.ts, derive the Ethereum-hub vote-collector satellite set from environments that expose both custom.wormhole.chainId and contracts.voteCollector, instead of relying on the hand-listed custom.governance.chainIds. Today Optimism is listed but has no wormhole config (so it's silently excluded from the votesCollected AND-check) and Moonbeam is listed but may not be a registered collector on the Ethereum multichainGovernor (which would pin votesCollected false). Separately, verify on-chain whether the Ethereum multichainGovernor has a vote collector registered for Wormhole chain 16 (Moonbeam). Add tests for (a) a declared satellite silently excluded by the filter and (b) a satellite with zero participation, so the AND-semantics edge cases are locked in.
```

**3. Stale "no-op" comments now misdescribe behavior** — `src/actions/governance/proposals/getProposals.ts:130`, `src/actions/governance/proposals/getProposal.ts:130`

Both comments claim that for cross-chain proposals `onChainData.state` is API-derived and `votesCollected` is always false, so the Queued-promotion branch is a no-op. After this PR, Ethereum-hub proposals get **real on-chain state** and `votesCollected` can be **true**, so that branch (≈ lines 148-154) now actively fires. A maintainer trusting the comment could delete it and silently revert the fix.

```text
Update the comments at src/actions/governance/proposals/getProposals.ts:130 and src/actions/governance/proposals/getProposal.ts:130. They claim the Queued-promotion post-processing is a no-op for cross-chain proposals because onChainData.state is API-derived and votesCollected is always false. After this PR, Ethereum-hub proposals receive real on-chain state and votesCollected can be true, so the branch is now live for them. Reword to reflect that it actively promotes Ethereum-hub proposals to Queued.
```

#### 🔵 Minor

**4. Two passes recompute `isLocal`/`homeEnv`/`isMultichain` independently** — `src/actions/governance/proposals/common.ts:378-380` vs `459-462`. The local-vs-foreign multichain heuristic (`isMultichainAware` vs `isMultichainProposal`) lives in two places; changing one pass but not the other yields a proposal classified multichain for state reads but single-chain for vote collection. Consider a single pre-pass producing `{ isLocal, homeEnv, isMultichain, proposalQuorum }` consumed by both.

**5. `state` and `proposalData` reads are now sequential** — `src/actions/governance/proposals/common.ts:392,406`. Previously a single `Promise.all`; the split (to attach independent `try/catch`) made them serial, ~2× RPC latency per proposal. `Promise.allSettled([...])` keeps both parallelism and per-read error handling.

**6. Decoupled `try/catch` can emit a hybrid record** — `src/actions/governance/proposals/common.ts:406`. If `state()` succeeds but `proposals()` rejects, `stateReadFailed` stays false, `proposalData` becomes null, and `eta` silently falls back to `votingEndTime + 86400` instead of the real on-chain eta.

**7. `getEnvironmentByChainId` duplication** — `src/actions/governance/proposals/common.ts:352`. `Object.values(publicEnvironments).find(e => e.chainId === …)` is duplicated at lines 300, 476, and 665. Extract a shared helper.

### Positive observations

- Per-read `try/catch` routed through `onError` with structured `{ source, chainId }` context matches the existing telemetry pattern and is a real improvement over the prior single swallow-all `console.warn`.
- The API-derived `state` fallback on read failure (so terminal proposals don't surface as `Pending(0)`) is a thoughtful degradation path.
- Comprehensive test additions in `common.test.ts`, including negative controls.
- Correctly scopes `legacyArtemisMaxId` to the local env and uses the targets-only heuristic for foreign envs.

### Suggestions

- Consider keeping `eta` as `bigint` until the UI boundary; `Number(proposalData[4])` plus raw tuple-index access is fragile if a governor ABI ever reorders fields. A named destructure would turn an ABI drift into a compile error rather than a silent wrong countdown.

```text
In src/actions/governance/proposals/common.ts, replace the raw tuple-index access proposalData[4] for eta with a named accessor or destructure, and keep eta as bigint until the point a number is required, so an ABI field-order divergence between the Moonbeam governor and the Ethereum multichainGovernor surfaces as a type error instead of a silent wrong eta.
```

---

apokusin:
## Follow-up review — after `8f936e1` + on-chain verification

### Resolved in the latest commit ✅

`8f936e1` cleanly addresses the earlier findings:

- **`chainIds` overload mis-routing Moonbeam's `homeEnvironment`** → `ethereum/custom.ts` reverted to `chainIds: []`; the satellite set is now *derived* via `getVoteCollectorSatellites`; new regression test *"Moonbeam homeEnvironment still resolves to Moonbeam"* (`ethereum/environment.test.ts`) locks it in. ✅
- **Stale "no-op" comments** in `getProposal.ts` / `getProposals.ts` → removed. ✅
- **Sequential `state`/`proposals` reads** → restored to concurrent via `Promise.allSettled`, keeping the independent per-read error handling. ✅
- **Duplicated env lookup** → extracted `getEnvironmentByChainId`, reused in `readCrossChainQuorums` + `resolveHomeEnv`. ✅

Typecheck clean; 35/35 governance + ethereum-env tests pass. I also confirmed the derivation is behavior-preserving for the Moonbeam-hub path (`getVoteCollectorSatellites(moonbeam)` = `{base}`, same as the old `[base, optimism]`→filtered).

### On-chain verification of the satellite set

I verified the Ethereum hub governor `0x8769B70ac7c93AF0e75de0D69877709B66d75838` (`MultichainGovernorV2`, deployed via the MIP-X58 Moonbeam→Ethereum migration) directly against mainnet:

```
$ cast call 0x8769…5838 "getAllTargetChains()(uint16[])"
[16, 30, 24]
  targetAddress(16) → 0xB8A798a50a7274A13449B7f2Dd6Df22faF2d40E5  (== SDK moonbeam voteCollector)
  targetAddress(30) → 0xe0278B32c627FF6fFbbe7de6A18Ade145603e949  (== SDK base voteCollector)
  targetAddress(24) → 0x3C968481BE3ba1a99fed5f73dB2Ff51151037738  (== SDK optimism voteCollector)
```

**Moonbeam (Wormhole 16) IS a registered collector** — so my earlier "needs on-chain verification" caveat is resolved, and the AND-check will *not* stick-false because of Moonbeam. ✅ But this same read surfaces a confirmed gap ↓

### Remaining issues

#### 🟡 1. Optimism is a registered Ethereum-hub collector on-chain, but the SDK silently drops it — `src/environments/definitions/optimism/custom.ts`

The governor registers **three** collection chains — 16 (Moonbeam), 30 (Base), **24 (Optimism)** — and `targetAddress(24)` is byte-for-byte the SDK's own `optimism` `voteCollector`. But `optimism/custom.ts` has **no `wormhole` block**, so `getVoteCollectorSatellites` (`common.ts:301`, which requires both `custom.wormhole.chainId` and `contracts.voteCollector`) excludes it. The derived satellite set is therefore `{Moonbeam, Base}` while the on-chain truth is `{Moonbeam, Base, Optimism}`. Consequence: `votesCollected` can flip `true` on Moonbeam+Base alone while Optimism's votes are still uncollected → the timeline shows **"Ready to Execute" prematurely**.

```text
In src/environments/definitions/optimism/custom.ts, add a `wormhole: { chainId: 24 }` block to the createCustomConfig call. The Ethereum hub MultichainGovernor (0x8769B70ac7c93AF0e75de0D69877709B66d75838) registers Optimism (Wormhole chain 24) as a vote-collection chain: getAllTargetChains() returns [16, 30, 24] and targetAddress(24) equals Optimism's voteCollector (0x3C968481BE3ba1a99fed5f73dB2Ff51151037738). Without wormhole.chainId, getVoteCollectorSatellites in src/actions/governance/proposals/common.ts silently excludes Optimism from the votesCollected AND-check, so a proposal can report votesCollected=true before Optimism's votes are collected. Add a test asserting getVoteCollectorSatellites for the Ethereum hub yields three satellites including Optimism.
```

#### 🟡 2. The "every satellite non-zero" gate is structurally fragile — `src/actions/governance/proposals/common.ts:524`

Verified against the governor's design: cross-chain tallies are bridged in only during a post-voting **24h `crossChainVoteCollectionPeriod`**, and a satellite where **nobody voted** reports `[0,0,0]` even after collection completes. So `votesCollectedChecks.every(...)` can pin `votesCollected` **false forever** for a low-participation proposal — re-creating the stuck timeline this PR set out to fix. Note that correctly fixing #1 (adding Optimism) makes this *more* likely, not less.

This path is also empirically unproven: the governor is ~8 days old, has exactly one proposal (**id 169, currently Active**), and has **never** completed a cross-chain collection (`CrossChainVoteCollected` event count = 0). So `votesCollected: true` has never actually occurred on this contract.

More robust: gate the timeline transition on the governor's own proposal state machine (`Active → CrossChainVoteCollection → Succeeded/Defeated`) or the collection-period deadline, rather than on all-satellites-non-zero tallies — which also makes the exact satellite set far less load-bearing.

```text
In src/actions/governance/proposals/common.ts, reconsider the votesCollected gate at line 524. It requires non-zero chainVoteCollectorVotes on every derived satellite (votesCollectedChecks.every(...)). The Ethereum MultichainGovernorV2 collects cross-chain votes only during a post-voting 24h crossChainVoteCollectionPeriod, and a satellite where nobody voted reports [0,0,0] permanently, so the AND-gate can pin votesCollected false forever for a low-participation proposal. Gate the timeline transition on the governor's proposal state (Active → CrossChainVoteCollection → Succeeded/Defeated) or the collection-period deadline instead of all-satellites-non-zero tallies.
```

#### 🔵 3. Minor (carried over, not blocking)

- Two passes still recompute `isLocal`/`homeEnv`/`isMultichain` independently (`common.ts:401` vs `487`) — drift risk if the multichain heuristic is changed in one but not the other.
- `Promise.allSettled` keeps independent error handling, so a `state`-ok / `proposals`-fail still yields a synthetic-eta hybrid record (`common.ts:451-466`): real on-chain state but `eta = votingEndTime + 86400` instead of the real on-chain eta.
- Optional: keep `eta` as `bigint` and read it by a named field rather than `proposalData[4]` (`common.ts:459`), so an ABI field-order divergence is a compile error rather than a silent wrong countdown.

### Note on PR #291

`#291` ("Fix moonbeam vote collector") is **orthogonal** to this PR despite the name — it restores Moonbeam voting *power* (xWELL `tokenVotes` passthrough, removes `RAW_WELL_MASKED_CHAINS`, deletes `votingPolicy.ts`) and touches `getUserVotingPowers.ts` / `getDelegates.ts`. It shares **zero files** with #290 and #290's code doesn't import `votingPolicy.ts`, so landing it doesn't change findings #1–#2.

---

apokusin (dismissed):

apokusin (commented):
## Review — Ethereum proposals timeline fix

The foreign-env resolution, `Promise.allSettled` split, RPC wiring, Wormhole IDs (16/30/24 = Moonbeam/Base/Optimism), and the `chainIds: []` rationale all check out, and deriving `votesCollected` from the governor state machine is the right call — the old per-satellite AND-gate genuinely was broken for low-participation proposals.

But there's one **🔴 Critical** issue: the new `votesCollected` definition interacts with the *unchanged* consumer promotion logic to mislabel terminal-failed proposals as `Queued`. Details inline. The unit tests lock in `votesCollected: true` for Canceled/Defeated but no end-to-end test asserts the rendered `ProposalState`, so it's currently green.

Non-blocking: a dead exported helper and a borderline `patch` vs `minor` changeset, also inline.

src/actions/governance/proposals/common.ts
```diff
@@ -393,91 +463,43 @@ export const getProposalsOnChainData = async (
...
+      // `state > MultichainVoteCollection`. Reading per-satellite tallies and
+      // AND-ing them would pin false forever for a satellite where no one
+      // voted (its entry stays [0,0,0] even after collection ends).
+      const votesCollected =
+        isMultichain &&
+        !stateReadFailed &&
+        state > MultichainProposalState.MultichainVoteCollection;
+
+      return {
+        state,
```
apokusin: **🔴 Critical — Defeated/Canceled proposals will render as `Queued` (with a fake timelock countdown).**

`votesCollected = state > MultichainProposalState.MultichainVoteCollection` is `true` for **Canceled(2), Defeated(3), Succeeded(4)** — the test at `common.test.ts:513-524` asserts exactly this. The returned `state` is the raw `MultichainProposalState`.

The consumers are unchanged (`getProposal.ts:143-149`, `getProposals.ts:141-149`):
```js
} else if (
  isMultichain && onChainData.votesCollected &&
  now > apiProposal.votingEndTime &&
  proposalState < ProposalState.Queued   // 5
) { proposalState = ProposalState.Queued; }
```
For a **Defeated** Ethereum-hub proposal: `state=3`, `votesCollected=true`, voting over, `3 < 5` → promoted to **Queued(5)**. Same for post-voting **Canceled(2)**. And since `proposals()` succeeds, the eta branch also sets `votingEndTime + 86400`, so it shows a running timelock + "Ready to Execute."

This is a **regression for the exact proposals this PR targets** — the old foreign path returned `votesCollected: false`, so Defeated/Canceled stayed correct. The promotion branch means "Succeeded → Queued"; the old AND-gate accidentally distinguished success from failure (a satellite with no voters → false), and `state > 1` removes that distinction while leaving the over-broad `< Queued` condition.

**Suggested prompt:**
```text
In getProposal.ts and getProposals.ts, the promote-to-Queued branch fires for any multichain proposal with votesCollected && proposalState < ProposalState.Queued. Since getProposalsOnChainData now sets votesCollected=true for Canceled/Defeated/Succeeded, defeated/canceled proposals get wrongly promoted. Normalize getProposalsOnChainData's returned state through MultichainProposalStateMapping (already imported) so it always returns a ProposalState, then gate the promotion on the specific awaiting-execution state rather than `< Queued`. Add a getProposals.test.ts case feeding {state: MultichainProposalState.Defeated, votesCollected: true} past votingEndTime and assert final proposal.state is Defeated, not Queued.
```

src/actions/governance/proposals/common.ts
```diff
@@ -393,91 +463,43 @@ export const getProposalsOnChainData = async (
...
+      // Cross-chain vote collection is the multichain governor's own state
+      // machine: votes can only be bridged in during the post-voting
+      // `crossChainVoteCollectionPeriod`, and once that window closes the state
+      // advances to Succeeded/Defeated. So "collection complete" is exactly
+      // `state > MultichainVoteCollection`. Reading per-satellite tallies and
+      // AND-ing them would pin false forever for a satellite where no one
+      // voted (its entry stays [0,0,0] even after collection ends).
+      const votesCollected =
+        isMultichain &&
+        !stateReadFailed &&
```
apokusin, line 475: **🟡 Major — returned `state` mixes two enum spaces.** On a successful read `state` is a `MultichainProposalState` (0–5); on `stateReadFailed` it's a `ProposalState` (0–9) from `deriveProposalStateFromApi`. Consumers read it unmapped. Beyond the Critical above, `MultichainVoteCollection(1)` surfaces as `ProposalState.Active(1)` rather than the Queued/"Vote Collection" phase `MultichainProposalStateMapping` defines. Normalizing the success branch through `MultichainProposalStateMapping` here would make both branches return `ProposalState` consistently — and is the natural home for the Critical fix too.

src/actions/governance/proposals/common.ts
```diff
@@ -283,6 +284,31 @@ const getLegacyArtemisMaxId = async (
...
+    (e) => e.chainId === chainId,
+  );
+
+/**
+ * Wormhole satellites the given hub talks to for cross-chain vote collection:
+ * every `publicEnvironments` entry other than the hub that exposes both
+ * `custom.wormhole.chainId` and `contracts.voteCollector`. Derived (not
+ * hand-listed) so the set stays consistent with the actual wiring.
+ */
+export const getVoteCollectorSatellites = (hubChainId: number): Environment[] =>
```
apokusin: **🔵 Minor (non-blocking) — dead export.** `getVoteCollectorSatellites` has no production caller — only `environment.test.ts` references it. Meanwhile `getExtendedProposalData` (~line 628) still derives its satellite set inline from `governance.chainIds` with a near-identical filter, so the doc comment's "derived... so the set stays consistent" promise isn't actually realized in the tally path. Either wire `getExtendedProposalData` to call this helper (removing the second source of truth), or drop the helper and fold its assertion into a direct test.

src/actions/governance/proposals/common.test.ts
```diff
@@ -288,80 +282,285 @@ describe("getProposalsOnChainData cross-chain skip", () => {
...
+    expect(data?.votesCollected).toBe(false);
+  });
+
+  test.each([
+    ["Canceled", MultichainProposalState.Canceled],
+    ["Defeated", MultichainProposalState.Defeated],
+    ["Succeeded", MultichainProposalState.Succeeded],
+    ["Executed", MultichainProposalState.Executed],
+  ])(
+    "flips true once governor state advances past MultichainVoteCollection (%s)",
```
apokusin, line 568: **🟡 Major — test gap that hides the Critical.** This correctly asserts `getProposalsOnChainData` returns `votesCollected: true` for Canceled/Defeated, but no test in `getProposal.test.ts`/`getProposals.test.ts` asserts the *rendered* `ProposalState` for a foreign multichain proposal in a terminal state. That's exactly why the Defeated→Queued bug stays green. Adding an end-to-end case (state=Defeated, votesCollected=true, now past votingEndTime → expect Defeated) would surface it.

.changeset/ethereum-proposal-timeline-data.md
```diff
@@ -0,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
```
apokusin: **🔵 Minor — changeset level.** `patch` is defensible (neither new helper is re-exported from `src/index.ts`), but the returned `state`/`votesCollected`/`eta` for foreign proposals change observably for consumers reading those fields — `minor` would signal that more honestly.

apokusin (commented):
## Re-review of `f67f4f1` — all prior findings resolved ✅

The `normalize multichain state and tighten Queued promotion` commit addresses every point from the previous review. Verified by tracing the logic and running the suite (`common` / `getProposal` / `getProposals` / `ethereum environment` → **69 passed**).

**🔴 Critical (Defeated/Canceled → Queued) — fixed.** The promotion gate is tightened from `proposalState < ProposalState.Queued` to `proposalState === ProposalState.Succeeded` in both `getProposal.ts:147` and `getProposals.ts:147`. Walking the states for a post-voting foreign multichain proposal: Canceled→Canceled, Defeated→Defeated, Succeeded→Queued, Executed→Executed. New `test.each` cases in both consumer test files assert Defeated/Canceled stay terminal despite `votesCollected: true`.

**🟡 Major (enum mixing) — fixed.** `getProposalsOnChainData` now normalizes the read-success multichain branch through `MultichainProposalStateMapping` (`common.ts:487`), so the returned `state` is always `ProposalState`. `votesCollected` is correctly computed against the raw governor enum *before* normalization (comment calls this out). Tests cover VoteCollection(1)→Queued(5), Executed(5)→Executed(7), and Defeated identity.

**🟡 Major (test gap) — fixed.** End-to-end `getProposal`/`getProposals` tests now assert the rendered `ProposalState` for terminal foreign proposals — exactly the coverage that was missing.

**🔵 Minor (dead export) — fixed.** `getVoteCollectorSatellites` removed; `environment.test.ts` inlines the satellite filter directly. No dangling references remain.

**🔵 Minor (changeset) — fixed.** Bumped `patch` → `minor`, and the changeset now documents the normalization + gate-tightening.

Nothing new of concern. The only nit — the `as Record<number, ProposalState>` cast at `common.ts:487` is a type assertion (which CLAUDE.md discourages in source) — is consistent with the existing `as Environment[]` / `as any` casts already in this file, so it's not worth churning. LGTM.

apokusin (approved):

PR #291, opened 2026-05-28
Fix moonbeam vote collector

bprofiro:
## Summary                                                                                                                                                                                                                                     
                                                                                                      
  The Moonbeam `MoonwellViews` implementation was upgraded on-chain to read xWELL for `tokenVotes`, matching the pattern Base and Optimism already use. The temporary `RAW_WELL_MASKED_CHAINS` workaround in `getUserVotingPowers` and `getDelegates` is no longer needed — both actions now return the views response unchanged on every multigov chain, so `totalDelegated` once again reflects the user's full xWELL + stkWELL + claims voting power on Moonbeam.
                                                                                                                                                                                                                                                 
  ## Changes                                                                                                                                                                                                                                     
                                                            
  - Delete `src/actions/governance/votingPolicy.ts` (sole consumers updated below).                                                                                                                                                              
  - `getUserVotingPowers.ts`: drop the per-chain mask; `tokenVotes` flows through from the views read.
  - `getDelegates.ts`: drop the `eligibleTokenVotes` branch; `totalVotes` is an unconditional sum.                                                                                                                                               
  - Test: replace the "Moonbeam zeroes out raw WELL" case with a Moonbeam pass-through regression test that pins the new behavior.

### DIFF
--- a/.changeset/moonbeam-restore-token-votes-passthrough.md
+++ b/.changeset/moonbeam-restore-token-votes-passthrough.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Remove the Moonbeam-specific `tokenVotes` mask from `getUserVotingPowers` and `getDelegates`. The Moonbeam `MoonwellViews` implementation was upgraded on-chain to read xWELL for `tokenVotes` (matching Base and Optimism), so the temporary `RAW_WELL_MASKED_CHAINS` workaround is no longer needed. Both actions now return the views response unchanged on Moonbeam, the same as every other chain, so `totalDelegated` once again reflects the user's full xWELL + stkWELL + claims voting power on Moonbeam instead of zeroing the xWELL contribution.
--- a/src/actions/governance/getDelegates.ts
+++ b/src/actions/governance/getDelegates.ts
@@ -9,7 +9,6 @@ import * as logger from "../../logger/console.js";
 import type { Delegate } from "../../types/delegate.js";
 import { getWithRetry } from "../axiosWithRetry.js";
 import { fetchAllVoters } from "./governor-api-client.js";
-import { RAW_WELL_MASKED_CHAINS } from "./votingPolicy.js";
 
 export type GetDelegatesReturnType = Promise<Delegate[]>;
 
@@ -65,17 +64,10 @@ export async function getDelegates(
         const { claimsVotes, stakingVotes, tokenVotes } =
           userVotingPowers[reduceIndex]!;
 
-        // Raw WELL (tokenVotes) is no longer an eligible voting source on
-        // Moonbeam — see comment in getUserVotingPowers.ts. Mask it here so
-        // delegate rankings reflect the new policy.
-        const eligibleTokenVotes = RAW_WELL_MASKED_CHAINS.has(env.chainId)
-          ? 0n
-          : tokenVotes.delegatedVotingPower;
-
         const totalVotes =
           claimsVotes.delegatedVotingPower +
           stakingVotes.delegatedVotingPower +
-          eligibleTokenVotes;
+          tokenVotes.delegatedVotingPower;
 
         votingPower[env.chainId] = Number(totalVotes / BigInt(10 ** 18));
       });
--- a/src/actions/governance/getUserVotingPowers.test.ts
+++ b/src/actions/governance/getUserVotingPowers.test.ts
@@ -1,4 +1,4 @@
-import { type Address, zeroAddress } from "viem";
+import type { Address } from "viem";
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { testClient } from "../../../test/client.js";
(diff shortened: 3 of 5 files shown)


apokusin:
## Code Review

### Summary

This PR removes the Moonbeam `tokenVotes` mask — deletes `votingPolicy.ts` (`RAW_WELL_MASKED_CHAINS`) and drops the per-chain mask in `getUserVotingPowers` and `getDelegates`, so `tokenVotes` flows through unchanged on Moonbeam. The **code change itself is clean** (see positives). The concern is its stated premise: *"the Moonbeam `MoonwellViews` implementation was upgraded on-chain to read xWELL for `tokenVotes`."* I checked the deployed contract the SDK actually targets, and **that upgrade is not live there** — so as of now, merging this re-introduces the raw-WELL over-count that the mask (#288) was added to fix.

### Issues

#### 🔴 Critical — the on-chain premise isn't true for the views address the SDK uses

`getUserVotingPowers` reads `env.contracts.views.getUserVotingPower(...)`, and Moonbeam's `views` is `0xe76C8B8706faC85a8Fbdcac3C42e3E7823c73994` (`moonbeam/contracts.ts`). I verified it against Moonbeam mainnet:

```
proxy 0xe76C8B…3994 → impl 0xdec80bb934397575594e91970b37baf65f5b21be  (MoonwellViewsV1Moonbeam, verified source)

cast call 0xe76C8B…3994 "governanceToken()(address)"  → 0x511aB53F793683763E5a8829738301368a2411E3   (legacy WELL)
cast call 0xe76C8B…3994 "safetyModule()(address)"     → 0x8568A675384d761f36eC269D695d6Ce4423cfaB1   (stkWELL ✓)

legacy WELL 0x511a…11E3  totalSupply = 5e27   (full 5B supply)
xWELL       0xA885…96AE  totalSupply = 1.906e20 on Moonbeam (~190 tokens)
```

- The verified source shows `getUserVotingPower` → `getUserTokensVotingPower` reads the `governanceToken` state variable, and `governanceToken()` returns **legacy WELL `0x511a…`**, *not* xWELL (`0xA885…`). The implementation is still `MoonwellViewsV1Moonbeam`.
- The supply disparity is corroborating: if the views read xWELL, nearly every Moonbeam user's `tokenVotes` would be ~0 (only ~190 xWELL exist on Moonbeam), which is clearly not the deployed behavior.

So the Moonbeam views still computes `tokenVotes` from legacy/raw WELL. Removing the mask makes `getUserVotingPowers.totalDelegated` and `getDelegates` voting power **count raw WELL again on Moonbeam** — exactly the over-count `RAW_WELL_MASKED_CHAINS` (#288) was added to suppress. Consumers (frontend delegate rankings, voting-power display) would show inflated Moonbeam numbers.

This is very likely the *"work we're waiting on to land"*: the change is correct **only after** the Moonbeam views upgrade is actually deployed. Recommendation — gate the merge on verifying the upgrade is live:

```text
Before merging PR #291, verify the Moonbeam MoonwellViews upgrade is actually deployed. Run:
  cast call 0xe76C8B8706faC85a8Fbdcac3C42e3E7823c73994 "governanceToken()(address)" --rpc-url https://rpc.api.moonbeam.network
It must return the xWELL address 0xA88594D404727625A9437C3f886C7643872296AE (currently it returns legacy WELL 0x511aB53F793683763E5a8829738301368a2411E3). If the upgrade instead deploys a NEW views contract at a different address, update src/environments/definitions/moonbeam/contracts.ts `views` to that address as part of this PR. Do not merge while getUserVotingPower's tokenVotes is still backed by legacy WELL, or Moonbeam totalDelegated will over-count raw WELL again.
```

*(Caveat: I could not complete a live `getUserVotingPower` vs xWELL/legacy-WELL holder comparison — Moonscan blocks automated holder lookups and my log-scan for a delegate didn't land. But `governanceToken()` + the verified source are authoritative on which token `tokenVotes` reads, so the conclusion holds independent of that cross-check.)*

### Positive observations

- **Clean removal.** No dangling references to `votingPolicy` / `RAW_WELL_MASKED_CHAINS` / `ZERO_TOKEN_VOTES` remain, and `votingPolicy.ts` was never re-exported from a public `index.ts`, so deleting it is **not** a breaking API change.
- The mask is removed **symmetrically** in both consumers (`getUserVotingPowers.ts` and `getDelegates.ts`), so they stay consistent.
- Changeset present; the test is updated from the "Moonbeam zeroes out raw WELL" case to a pass-through regression test that pins the new behavior in both directions.

### Suggestions

- The updated test mocks the views response, so it passes regardless of what the real contract returns — it can't catch the premise being wrong. Consider asserting the gating fact directly, or add a comment noting the behavior depends on the on-chain `governanceToken` being xWELL.

```text
In getUserVotingPowers.test.ts, add a comment (or a lightweight integration assertion) documenting that the Moonbeam pass-through is only correct once the on-chain MoonwellViews governanceToken() returns xWELL (0xA88594…), since the unit test mocks the views read and would pass even if the real contract still returned legacy WELL.
```

---

apokusin (approved):

PR #293, opened 2026-05-29
Add mainnet support for comptroller and markets to moonwell

bprofiro:
## Summary

Wires Moonwell Core lending on Ethereum mainnet (chain 1) into the SDK so consumers can read the 4 launch markets — **mWETH, mUSDC, mUSDT, mcbBTC** — sourced from the Lunar Indexer.

## Changes

- `environments/definitions/ethereum/tokens.ts` — add WETH / USDT / cbBTC underlyings + the 4 mTokens
- `environments/definitions/ethereum/core-markets.ts` *(new)* — `createMarketConfig` mapping each mToken to its underlying
- `environments/definitions/ethereum/environment.ts` — wire `markets` into the env factory + return type
- `environments/index.ts` — add the `EthereumEnvironment` branch to `MarketsType<>` (was `undefined`, now resolves to the 4 market keys)

Companion to indexer PR [moonwell-fi/lunar-indexer#52](https://github.com/moonwell-fi/lunar-indexer/pull/52).

## Effect

- `getMarkets()` returns Ethereum entries where it previously returned nothing for chain 1
- `keyof MarketsType<EthereumEnvironment>` resolves to `"MOONWELL_WETH" | "MOONWELL_USDC" | "MOONWELL_USDT" | "MOONWELL_cbBTC"`
- Snapshots / positions / staking actions already chain-multiplex through `lunarIndexerUrl`, so no further wiring needed

## Verification

- `pnpm typecheck` ✅
- `pnpm lint` ✅
- `pnpm test` — **508 tests pass** ✅
- Live indexer check: `GET /api/v1/core/markets/1` returns the 4 expected markets at the addresses registered here


src/environments/definitions/ethereum/core-markets.ts
```diff
@@ -0,0 +1,24 @@
+import { createMarketConfig } from "../../types/config.js";
+import { tokens } from "./tokens.js";
+
+export const markets = createMarketConfig({
```
apokusin, line 4: **Suggestion (non-blocking):** Consider adding a test that locks in these markets. Nothing currently asserts the 4 Core markets are registered, yet the changeset claims `getMarkets()` now returns chainId=1 entries and `MarketsType<EthereumEnvironment>` resolves to concrete keys.

A future refactor reverting `markets: markets` back to `markets: {}`, dropping a market, or transposing an mToken address would keep CI green. `src/environments/definitions/ethereum/environment.test.ts` is already in the `test/vitest.config.ts` include array, so it's near-zero friction:

```text
In src/environments/definitions/ethereum/environment.test.ts, add a test that imports the ethereum environment and asserts all 4 core markets (MOONWELL_WETH, MOONWELL_USDC, MOONWELL_USDT, MOONWELL_cbBTC) are present in environment.markets with the correct underlyingToken keys (WETH, USDC, USDT, cbBTC).
```

src/environments/definitions/ethereum/environment.ts
```diff
@@ -35,11 +42,17 @@ const createEnvironment = (
...
       lunarIndexerUrl || "https://lunar-services-worker.moonwell.workers.dev",
     tokens,
-    markets: {},
+    markets,
     vaults: {},
     morphoMarkets: {},
     contracts,
     custom,
-  }) as Environment<typeof tokens, {}, {}, typeof contracts, typeof custom>;
+  }) as Environment<
```
apokusin: **Suggestion (non-blocking, minor):** This trailing `as Environment<...>` cast is redundant — the function already has an explicit return-type annotation and `createEnvironmentConfig` casts internally. Removing it typechecks clean and aligns with CLAUDE.md's no-`as`-casts rule.

Pre-existing pattern (Base does the same), so not a regression — but since this PR is editing the line, dropping the trailing cast is a free cleanup. Keeping it means the 5-type-param tuple is written twice and must be kept in sync if vaults/morphoMarkets are added later.

apokusin (commented):
Reviewed the Ethereum Core markets addition — clean, additive change. Underlying addresses, mToken decimals, `MarketsType` ordering, and chain routing all verified correct. Two non-blocking suggestions below.

lyoungblood:
## Code Review — Ethereum Core markets

**Verdict: clean and verified correct.** This PR registers the 4 launch Moonwell Core markets on Ethereum mainnet (mWETH, mUSDC, mUSDT, mcbBTC) plus their underlyings, and wires `MarketsType<EthereumEnvironment>` to the new keys. No functional bugs found — the surviving items are minor consistency/coverage nits.

### What I verified (on-chain, not just by reading)

- **All 4 mToken addresses, symbols, decimals (8), and `underlying()` targets match the config** — confirmed via `eth_call` to mainnet. Underlying addresses/decimals (WETH 18, USDC 6, USDT 6, cbBTC 8) are correct.
- **The Lunar indexer returns exactly these 4 markets for `chainId=1`** (the PR's primary path) — confirmed against the live endpoint; `findMarketByAddress` matches all 4.
- **Ethereum's `views` contract is fully functional for core reads** — `getAllMarketsInfo()` returns the 4 markets and the per-user reads (`getUserBalances`/`getUserBorrowsBalances`/`getUserMarketsMemberships`) all return data without reverting. So `getMarkets` / `getUserPositions` / `getUserRewards` work on both the Lunar and on-chain fallback paths. (This refuted an early hypothesis that the views reverts and would surface markets as paused/$0 via the mToken fallback — it does not revert.)
- **Types are sound** — `pnpm typecheck` passes, the `MarketsType` branch ordering is safe (environments don't structurally overlap), and the wiring mirrors `base`'s exact annotation + cast pattern. No existing test breaks.
- The missing comptroller (despite the branch name) is a **non-issue**: no SDK action references `contracts.comptroller`.

### Issues

All 🔵 **Minor** — none blocking.

**🔵 `core-markets.ts:7` — ETH market keyed differently from other chains.** Ethereum exposes its ETH market as `MOONWELL_WETH` / underlying `WETH`, whereas `base` and `optimism` use `MOONWELL_ETH` / underlying `ETH` (native). A consumer aggregating ETH markets across chains by key (`MOONWELL_ETH`) or by `underlyingToken.symbol === "ETH"` would silently omit Ethereum's market. This is **defensible by design** — on-chain `mWETH.underlying()` genuinely is WETH (an ERC-20; no native CEther market on mainnet) — and no internal SDK consumer is affected (no hardcoded `MOONWELL_ETH`; the only `symbol === "ETH"` checks are Morpho-only in `getMarketSnapshots.ts`, and Ethereum has no Morpho markets). Worth a deliberate confirm rather than leaving it implicit.

**🔵 `tokens.ts:23` — USDT `name` inconsistency.** `"Tether USD"` here vs `"Tether"` on optimism/moonriver. Display-only (matching is by `address.toLowerCase()`). Note `"Tether USD"` is the canonical on-chain `name()`, so the cleaner fix is arguably aligning the other chains.

**🔵 `tokens.ts:29` — cbBTC `name` inconsistency.** `"Coinbase Wrapped BTC"` here vs `"Coinbase Bitcoin"` on base. Same class — display-only, and Ethereum's value is the canonical on-chain one.

**🔵 `core-markets.ts:4` — no test asserts the new markets.** `ethereum/environment.test.ts` (the one CI-gated per-chain test) asserts nothing about `markets`. Lower value than usual since `underlyingToken`/`marketToken` are typed as `keyof tokens` (key typos are already compile errors), but a small registration assertion would still help. Separately, the three all-environment iteration tests (`getMarket`/`getUserPosition`/`getUserReward.test.ts`) each gain ~8 live-mainnet-RPC cases on a full local run — but none is in the vitest `include` array, so CI and the quality gate are unaffected.

### Latent note (not currently triggered)

No environment lists `chainId 1` in `custom.governance.chainIds`, so the "home" environment for Ethereum resolves to itself, and `getNativeTokenPrice()` on its views returns `0`. Today every Ethereum market incentive is WELL (priced via Base's oracle) with zero emissions, so there's no wrong output — but if a future Ethereum market gains a **native-ETH** reward incentive, its reward USD/APR would compute as 0. One to keep in mind as markets/incentives grow.

### Positive observations

- Config exactly mirrors the established `createMarketConfig`/`createTokenConfig` pattern; the explicit `Environment<...>` annotation + cast matches `base` precisely.
- Correct minor Changeset included, accurately describing the public-type-surface change.
- Lowercase addresses are safe here — both `findMarketByAddress` and `findTokenByAddress` compare with `.toLowerCase()` on both sides.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

apokusin (approved):

PR #297, opened 2026-06-01
fix(ethereum): wire Unitroller as contracts.comptroller

bprofiro:
Adds the Moonwell Ethereum Unitroller (0xdec80bb934397575594e91970b37baf65f5b21be) to the Ethereum env so consumers can call enterMarkets / exitMarket on the 4 Core markets. Without it, the frontend's "Enable as collateral" modal built its useTransaction against an undefined contract address and silently no-op'd on confirm (no wallet prompt, no toast, no console error).

### DIFF
--- a/.changeset/ethereum-comptroller.md
+++ b/.changeset/ethereum-comptroller.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Wire the Ethereum mainnet Unitroller (`0xdec80bb934397575594e91970b37baf65f5b21be`) as `contracts.comptroller` on the Ethereum environment. Consumers can now call `enterMarkets` / `exitMarket` for the 4 Core markets — without this, the frontend's "Enable as collateral" modal built its `useTransaction` against an undefined contract address, which caused the confirm button to silently no-op (no wallet prompt, no toast, no console error).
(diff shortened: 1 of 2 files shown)


bprofiro:
The `audit.critical: 0 → 1` regression on the quality gate is unrelated to this PR. The branch only touches:

- `src/environments/definitions/ethereum/contracts.ts` (+6 lines)
- `.changeset/ethereum-comptroller.md` (new, +5 lines)

Neither `package.json` nor `pnpm-lock.yaml` is modified:

    $ git diff --stat main...HEAD
     .changeset/ethereum-comptroller.md                 | 5 +++++
     src/environments/definitions/ethereum/contracts.ts | 6 ++++++

The new critical advisory was disclosed by the npm registry against an existing dep on `main` since `quality-gate/baseline.json` was last re-captured — it would fail the gate on any PR opened today, including no-op ones.

Will be addressed in a follow-up PR that upgrades the offending package (or pins/overrides it) and re-captures the baseline. Per [CLAUDE.md](../blob/main/CLAUDE.md) we deliberately don't edit `baseline.json` to bypass the gate from inside an unrelated PR.

apokusin (approved):
lgtm, need confirmation that the address is correct

bprofiro:
it's the same one we're using on indexer. I've also already tested on the frontend: [0xdec80bb934397575594e91970b37baf65f5b21be](https://github.com/moonwell-fi/lunar-indexer/pull/52/changes#diff-7ec41dc94b8d579cf54c8163397268deedc6b9543bb899d7cca7c9d3b6acedbe)

PR #299, opened 2026-06-04
fix(governance): hub-aware classification, convergent block search, per-chain voting-power isolation

apokusin:
## Summary

Root-cause fixes for the **proposal-171 "cannot vote" incident** (Jun 3) — the frontend hotfix (moonwell-frontend-v2-react #1051) worked around these; this PR fixes them at the source. Tracks [MOO-352](https://linear.app/moonwell/issue/MOO-352), [MOO-353](https://linear.app/moonwell/issue/MOO-353), [MOO-354](https://linear.app/moonwell/issue/MOO-354) under incident parent [MOO-351](https://linear.app/moonwell/issue/MOO-351).

### 1. Hub-aware multichain classification (MOO-354)
`getProposal`/`getProposals` set `proposal.multichain` via the targets-only Wormhole-bridge heuristic — hub-local proposals (171: single Ethereum-side target, no bridge) classified as single-chain, routing frontend votes into a dead legacy-governor branch. New canonical `classifyProposalMultichain` (also exported: `isMultichainHomeChain`): **a proposal homed on a hub chain with no legacy governor is multichain by construction**, plus the existing bridge-target and Artemis-cutoff checks. Now drives `getProposal`, `getProposals`, and `getProposalsOnChainData` state-read routing so paths can't drift. `isMultichainAware` deprecated.

### 2. `getBlockNumberAtTimestamp` binary-search completion (MOO-352)
The 8-iteration interpolation search silently returned its unconverged lower bound. On Moonbeam (12s→6s block-time change) every probe overshoots, `lo` never leaves block 1, and recent timestamps resolved to **block 1 (Dec 2021)** — reverting historical voting-power reads. Now finishes with plain bisection: same fast path on linear chains, guaranteed convergence otherwise. An unconverged bound can never be returned.

### 3. `getUserVotingPowers` per-chain isolation (MOO-353)
Single `Promise.all` meant the Moonbeam failure erased voting power on every chain — hiding all vote buttons for all users. Now `Promise.allSettled`: failed chains report via `env.onError` (`source: "getUserVotingPowers"`) and are omitted. **Behavioral change** (in changeset): consumers get partial results instead of a rejection.

## Tests
- Classification matrix: hub-local (171 regression), bridged, legacy-Artemis, cutoff-past, cutoff-failed
- Moonbeam 12s→6s regime-change regression + randomized piecewise-regime invariant suite (`ts(result) ≤ target < ts(result+1)`)
- Per-chain isolation: views revert / block-resolution failure / all-chains-failed
- `votesCollected` gate test re-homed — the old test **pinned the incident's buggy assumption** (chainId-1 + local targets ⇒ non-multichain); hub-local proposals now assert the flip

## Verification (live mainnet, this branch built)
```
171 multichain: {"id":171,"votesCollected":false} | state: 1 (Active)   ← was undefined
169 multichain: {"id":169,"votesCollected":true}  | state: 7            ← unchanged
voting powers @171 snapshot: 1=1834651.54  8453=297.57  10=0.00  1284=0.00  ← all 4 chains, incl. Moonbeam
```
Note: `getUserVotingPowers` integration tests fail in my sandbox (no network in vitest workers) — verified identical failures on clean main; unit suites 11/11, all other touched suites 80/80.

## Release
Changeset: **minor** (new exports + allSettled behavioral change). After release, bump the frontend pin and reconcile the #1051 workarounds per [MOO-355](https://linear.app/moonwell/issue/MOO-355).


bprofiro:
### Re: Quality Gate `audit.critical` failure — unrelated to this PR

The gate is failing on `audit.critical` (0 → 1), but this is **not** introduced by the changes in this branch. It's audit-database drift, not a regression.

**Details:**
- This branch makes **zero** changes to `package.json` or `pnpm-lock.yaml` (`git diff main..HEAD -- package.json pnpm-lock.yaml` is empty).
- The flagged advisory is **GHSA-5xrq-8626-4rwp** in `vitest@2.1.9` ("When Vitest UI server is listening, an arbitrary file can be read and executed").
- It was **published after** the gate baseline was snapshotted, so `main` itself fails this same check today — it's not specific to these governance fixes.
- It is **dev-only**: the vulnerable code path is the Vitest UI server (`vitest --ui`), which we never run in CI and which is **not shipped** in the published SDK. It has no impact on package consumers.
- The only patched line is `>=4.1.0` (no 2.x/3.x patch exists), so remediation means a two-major `vitest` + `@vitest/coverage-v8` upgrade (2 → 4). That's out of scope for a governance PR and will be handled separately.

Per project policy I have **not** edited `quality-gate/baseline.json` to bypass the gate.

avp1598 (approved):

PR #306, opened 2026-06-15
fix(ethereum): add multiRewardDistributor for reward claims (MOO-413)

bprofiro:
## Summary

Adds the Ethereum `multiRewardDistributor` (`MRD_PROXY` = `0x60142B8d76FaC5b88cfB422Ba1aA905d2171851c`) to `src/environments/definitions/ethereum/contracts.ts`, mirroring the Base and Optimism configs. Closes [MOO-413](https://linear.app/moonwell/issue/MOO-413).

## Why (the real reason — not what the ticket says)

The ticket claimed the missing address would make reward **APRs display as 0%**. That's inaccurate — reward APRs come from the on-chain `views` contract / Lunar indexer, and the SDK never reads `multiRewardDistributor` directly.

The actual consumer is the **frontend** `useUserRewardsData.tsx`, which branches on this address to build the reward-**claim** transaction:

| `multiRewardDistributor` defined? | Claim routes to | ABI / fn |
| --- | --- | --- |
| Yes (Base, OP) | `comptroller.address` | `claimReward(holders, mTokens, …)` |
| No (Moonbeam/Moonriver) | `0x…0808` precompile | `batchAll(…)` |

Ethereum, lacking it, fell into the Moonbeam precompile path — a claim tx to `0x…0808` on Ethereum mainnet reverts. So once MIP-X59 makes WELL accrue on the four Ethereum Core markets (ETH, USDC, USDT, cbBTC), users would see rewards but **fail to claim** them. No frontend change needed — the frontend already branches on the value; this just supplies it.

## Verification

- **Address verified on-chain:** `Comptroller(0xdec80bb…21be).rewardDistributor()` on Ethereum mainnet returns exactly `0x60142B8d76FaC5b88cfB422Ba1aA905d2171851c`.
- `pnpm typecheck` clean · `pnpm lint` clean.
- Full suite green (33 files / 550 tests); `getUserRewards` live-RPC cases for Ethereum pass.
- Quality gate: **PASS** (no regression).

## Tests added

- `src/environments/definitions/ethereum/environment.test.ts` — locks both `env.config.contracts.multiRewardDistributor` (raw config) and `env.contracts.multiRewardDistributor.address` (wired viem contract), so dropping or mis-wiring the address fails CI. This closes the coverage gap that let the omission ship originally.

## Notes / follow-ups

[...]

PR #307, opened 2026-06-15
fix(config): require multiRewardDistributor for EVM comptroller chains

bprofiro:
## Summary

Closes the **root cause** behind [MOO-413](https://linear.app/moonwell/issue/MOO-413) / #306: the contracts config typed `multiRewardDistributor?` as optional, so an EVM comptroller chain could omit it with **no compile error** — which is exactly how Ethereum shipped without it and silently broke the frontend reward-claim flow (it falls back to the Moonbeam `0x…0808` precompile path, which reverts).

> ⚠️ **Stacked on #306** (base = `brenda/moo-413-…`). This change requires Ethereum's MRD to already be present, so it must merge after #306. Retarget to `main` once #306 lands.

## Change

`createContractsConfig` now intersects a conditional constraint onto its input:

```ts
type RequireMultiRewardDistributor<contracts> = contracts extends { comptroller: Address }
  ? contracts extends { governor: Address }
    ? unknown                              // Moonbeam/Moonriver: WELL precompile model, no MRD
    : { multiRewardDistributor: Address }  // EVM comptroller chain: MRD required
  : unknown;                               // no comptroller: not a lending env
```

**Discriminator:** `comptroller` present + `governor` absent ⟹ EVM comptroller chain ⟹ MRD required. Verified against all 8 definitions: Base/Optimism/Ethereum (require, and have it); Moonbeam/Moonriver (have `governor` → exempt); arbitrum/avalanche/polygon (no comptroller → exempt).

## Verification

- `pnpm typecheck` passes for all 8 current definitions.
- **Proof it catches the regression:** removing MRD from `ethereum/contracts.ts` now produces `error TS2741: Property 'multiRewardDistributor' is missing`.
- Moonbeam/Moonriver and the non-comptroller chains still compile.
- Full suite: 552 pass (34 files). Quality gate: PASS.

## Tests added

- `src/environments/contracts-invariants.test.ts` — runtime backstop iterating `publicEnvironments`: asserts every EVM comptroller env (comptroller, no governor) has a valid `multiRewardDistributor.address`. Covers envs built outside the factory and any `as`-cast bypass. Registered in the vitest include list.

## Reviewer notes (triaged)

- A type-design review flagged a possible `governor: undefined` bypass (🟡). **Verified not reproducible** under this repo's `exactOptionalPropertyTypes: true` — `governor: undefined` is itself a compile error (TS2375), so it can't take the exempt branch. No code change made (guarding an impossible case would add noise).
- The `governor`-presence discriminator is a structural proxy for "Moonbeam-family home chain." Sound for all current/foreseeable chains; if a future chain breaks the heuristic, an explicit `kind` tag would be the more durable design. Noted as a future option, not done now (disproportionate to this fix).

## Semver

Marked **minor** (breaking) per the repo's breaking-change hook: direct callers of the exported `createContractsConfig` that omit MRD on an EVM comptroller config will no longer compile. Internal definitions are all updated/valid.

### DIFF
--- a/.changeset/require-multirewarddistributor-evm-config.md
+++ b/.changeset/require-multirewarddistributor-evm-config.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": minor
+---
+
+Require `multiRewardDistributor` at the type level for EVM comptroller environments. **Type-breaking** for direct callers of `createContractsConfig`: a config with a `comptroller` and no `governor` that omits `multiRewardDistributor` no longer compiles. `createContractsConfig` now rejects a contracts config that defines a `comptroller` but no `governor` (i.e. Base/Optimism/Ethereum) unless it also defines `multiRewardDistributor` — closing the gap that let Ethereum ship without it and silently break reward claims (MOO-413). Moonbeam and Moonriver, which use the on-chain `governor` + WELL precompile reward model, are unaffected and correctly continue to omit it. A runtime invariant test across `publicEnvironments` backstops the type guard.
--- a/src/environments/contracts-invariants.test.ts
+++ b/src/environments/contracts-invariants.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, test } from "vitest";
+import { publicEnvironments } from "./index.js";
+
+// Runtime companion to the `createContractsConfig` compile-time guard in
+// types/config.ts. Every EVM comptroller environment — one that has a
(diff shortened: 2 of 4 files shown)


bprofiro:
Closing — not needed. The fix lives entirely in #306 (adds Ethereum's `multiRewardDistributor`, with an env test locking it). Making the config type *require* MRD is a breaking change to `createContractsConfig` that isn't necessary to solve MOO-413.

PR #314, opened 2026-07-03
Remove dead Ponder indexer fetches from snapshot actions

bprofiro:
## What

Removes the Ponder indexer fetch paths from four SDK actions:

- `getStakingSnapshots`
- `getCirculatingSupplySnapshots`
- `getMarketSnapshots` (core path)
- `getUserPositionSnapshots`

When an environment has no `lunarIndexerUrl` configured (only Moonriver), these actions now return `[]` immediately instead of querying Ponder. Net **−600 lines**.

## Why

The Ponder API (`ponder.moonwell.fi`) is decommissioned and returns HTTP 502, so these paths were dead code: every Moonriver call already failed inside a catch block and returned an empty array. Observable behavior is unchanged — the doomed network round-trip is simply gone. Lunar-indexer support for Moonriver snapshots is not planned (staking/circulating-supply were removed from the UI for Moonriver, market charts no longer render for deprecated markets, and per Luke user-position snapshots aren't needed there).

## Non-breaking by design

No public type changes. The deprecated `Environment.indexerUrl` field and the Moonriver environment's Ponder URL default are **deliberately kept** — nothing reads them anymore. Their removal is a breaking change tracked separately in [MOO-537](https://linear.app/moonwell/issue/MOO-537/remove-deprecated-indexerurl-ponder-field-from-sdk-public-types).

## Tests

Each deleted "Ponder path" suite is replaced with a behavioral test pinning the new contract: the action resolves to exactly `[]` (never `undefined`), makes no HTTP call (axios spies), and does not invoke `environment.onError` (guards against the empty result coming from the Lunar catch path instead of the intentional branch). The surviving Lunar paths and their tests are untouched.

## Reviewer notes

- `postWithRetry`/`getWithRetry` in `axiosWithRetry.ts` are kept — still used by the Snapshot-vote and Lunar circulating-supply code.
- One stale public JSDoc on `getUserPositionSnapshots` (describing the removed start-of-day filtering) was updated to match the Lunar-only behavior.
- Verified locally: lint, typecheck, all 4 test suites (156 tests), `pnpm build && pnpm test:build` (publint + attw), and the quality gate — all gated metrics hold or improve (coverage +1pp from deleting dead code). Baseline intentionally not re-captured in this PR.
- Changeset: patch bump.

### DIFF
--- a/.changeset/brave-mirrors-tell.md
+++ b/.changeset/brave-mirrors-tell.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Removed the dead Ponder indexer fetches. `getStakingSnapshots`, `getCirculatingSupplySnapshots`, `getMarketSnapshots` (core), and `getUserPositionSnapshots` now return `[]` for environments without a `lunarIndexerUrl` (Moonriver) instead of querying the decommissioned Ponder API. No public type changes — the deprecated `Environment.indexerUrl` field remains for now and its removal is tracked separately (MOO-537).
--- a/src/actions/core/markets/getMarketSnapshots.test.ts
+++ b/src/actions/core/markets/getMarketSnapshots.test.ts
@@ -182,7 +182,7 @@ describe("trimLeadingEmptySnapshots (unit)", () => {
   });
 
   test("trims leading zero snapshots (descending order)", () => {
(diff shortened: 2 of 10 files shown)


bprofiro:
Superseded by #315 (same commit, branch renamed to match Linear MOO-539). GitHub auto-closed this PR when the head branch was renamed.

PR #315, opened 2026-07-03
Remove dead Ponder indexer fetches from snapshot actions

bprofiro:
## What

Removes the Ponder indexer fetch paths from four SDK actions:

- `getStakingSnapshots`
- `getCirculatingSupplySnapshots`
- `getMarketSnapshots` (core path)
- `getUserPositionSnapshots`

When an environment has no `lunarIndexerUrl` configured (only Moonriver), these actions now return `[]` immediately instead of querying Ponder. Net **−600 lines**.

## Why

The Ponder API (`ponder.moonwell.fi`) is decommissioned and returns HTTP 502, so these paths were dead code: every Moonriver call already failed inside a catch block and returned an empty array. Observable behavior is unchanged — the doomed network round-trip is simply gone. Lunar-indexer support for Moonriver snapshots is not planned (staking/circulating-supply were removed from the UI for Moonriver, market charts no longer render for deprecated markets, and per Luke user-position snapshots aren't needed there).

## Non-breaking by design

No public type changes. The deprecated `Environment.indexerUrl` field and the Moonriver environment's Ponder URL default are **deliberately kept** — nothing reads them anymore. Their removal is a breaking change tracked separately in [MOO-537](https://linear.app/moonwell/issue/MOO-537/remove-deprecated-indexerurl-ponder-field-from-sdk-public-types).

## Tests

Each deleted "Ponder path" suite is replaced with a behavioral test pinning the new contract: the action resolves to exactly `[]` (never `undefined`), makes no HTTP call (axios spies), and does not invoke `environment.onError` (guards against the empty result coming from the Lunar catch path instead of the intentional branch). The surviving Lunar paths and their tests are untouched.

## Reviewer notes

- `postWithRetry`/`getWithRetry` in `axiosWithRetry.ts` are kept — still used by the Snapshot-vote and Lunar circulating-supply code.
- One stale public JSDoc on `getUserPositionSnapshots` (describing the removed start-of-day filtering) was updated to match the Lunar-only behavior.
- Verified locally: lint, typecheck, all 4 test suites (156 tests), `pnpm build && pnpm test:build` (publint + attw), and the quality gate — all gated metrics hold or improve (coverage +1pp from deleting dead code). Baseline intentionally not re-captured in this PR.
- Changeset: patch bump.

Closes MOO-539.

### DIFF
--- a/.changeset/brave-mirrors-tell.md
+++ b/.changeset/brave-mirrors-tell.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Removed the dead Ponder indexer fetches. `getStakingSnapshots`, `getCirculatingSupplySnapshots`, `getMarketSnapshots` (core), and `getUserPositionSnapshots` now return `[]` for environments without a `lunarIndexerUrl` (Moonriver) instead of querying the decommissioned Ponder API. No public type changes — the deprecated `Environment.indexerUrl` field remains for now and its removal is tracked separately (MOO-537).
--- a/src/actions/core/markets/getMarketSnapshots.test.ts
+++ b/src/actions/core/markets/getMarketSnapshots.test.ts
@@ -182,7 +182,7 @@ describe("trimLeadingEmptySnapshots (unit)", () => {
   });
 
   test("trims leading zero snapshots (descending order)", () => {
-    // Ponder returns newest-first; function must still work
+    // The indexer may return newest-first; function must still work
     const snapshots = [
       makeMarketSnapshot(DAY2, 100, 50),
       makeMarketSnapshot(DAY1, 0, 0),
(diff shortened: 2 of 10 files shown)


apokusin (approved):
## Review

**Verdict: Approve.** Clean, well-scoped dead-code deletion (+44/−640 across 10 files). I independently verified the PR's claims at head commit `240de53`: all 4 affected test suites pass (154 tests), `tsc --noEmit` is clean, and `biome check` reports nothing.

### Correctness — verified

- **No orphaned code**: `postWithRetry`/`getWithRetry` are genuinely still used (11 files, including this PR's surviving Lunar path in `getCirculatingSupplySnapshots` plus governance/beam/morpho actions). `isStartOfDay` remains imported in `getMarketSnapshots.ts` because the Lunar path still uses it (`getMarketSnapshots.ts:376`). `calculateTimeRange` in `getStakingSnapshots.ts` is still used by the Lunar fetcher.
- **No remaining readers of `environment.indexerUrl`**: the only surviving non-test references are the config plumbing (`src/environments/types/config.ts:428`) and the Moonriver default (`environment.ts:26`) — both deliberately kept and tracked in MOO-537, as stated.
- **Tests guard the right thing**: asserting `onError` was *not* called is a nice touch — it distinguishes the intentional early-return from the Lunar catch path silently swallowing an error.

### Notes (minor, none blocking)

- **One observable change the PR body slightly understates**: previously a Moonriver call to `getMarketSnapshots`/`getUserPositionSnapshots` would fail against Ponder and invoke `environment.onError` (sources `market-snapshots-ponder` / `user-position-snapshots-ponder`). That callback now never fires. For anyone piping `onError` into Sentry this is a *reduction in noise* — desirable, but strictly a behavioral change, not just a removed round-trip. The new tests correctly pin it.
- **Test-count nit**: PR body says 156 tests; I count 154 across the 4 suites. Likely a stale count from before the final edit.
- **Small inconsistency**: the new `getCirculatingSupplySnapshots` and `getStakingSnapshots` tests don't add the `onError` spy that the market/user-position tests do (staking compensates by asserting `createLunarIndexerClient` isn't called). Fine to leave as-is.
- **Incidental hygiene win**: the deleted Ponder fetchers built GraphQL queries via string interpolation of addresses; removing them eliminates that (low-risk but ugly) injection surface.

### Conventions & release

- Changeset present with a patch bump — appropriate: no public type changes, and the removed path already returned `[]` in practice.
- JSDoc on `getUserPositionSnapshots` updated to drop the stale start-of-day-filtering claim — good catch of comment rot.

PR #318, opened 2026-07-13
MOO-587: Fix 10x-inflated Merkl staking rewards from repeated breakdown pages

james-saint:
## Summary

Staking rewards for the safety module were displayed at up to **10× their actual value** on Base (e.g. the app showed **7,648 WELL** to claim while Merkl showed **764.9 WELL** for `0x58b6920649d18b3d8162547e7ced2b3f46d7175e`).

**Root cause:** `getMerklRewardsData` paginates the Merkl `/rewards` endpoint via `breakdownPage`, appending each page's breakdowns into the page‑0 data. But Merkl **clamps an out‑of‑range `breakdownPage` back to page 0** instead of returning an empty page, so the same breakdowns were re‑appended on every page. With `MAX_BREAKDOWN_PAGES = 10`, each reward breakdown was counted up to 10 times — exactly the reported 10× inflation. `getUserStakingInfo` then sums `amount − claimed` over every (duplicated) breakdown, so the claimable total was multiplied.

I confirmed this live against the Merkl proxy for the example address: `breakdownPage=0`, `1`, and `2` all return the **same 8 breakdowns**, and the reward's top‑level `amount` equals the sum of the page‑0 breakdowns (proving page 0 is already complete).

## Fix

- Seed a `seenBreakdowns` set from page 0 and, on later pages, only append breakdowns whose content key is not already seen.
- Break out of the pagination loop as soon as a page contributes **no new** breakdowns (the clamp‑to‑page‑0 signal), which also avoids 9 redundant fetches.
- Genuinely paginated, distinct breakdowns are still merged; distributions that legitimately share a `campaignId` but differ in amount are preserved (keyed on full breakdown content).

Scoped entirely to the SDK — the frontend reads `userStakingInfo.pendingRewards.value` directly and needs no change; it picks up the fix via the SDK.

## Test evidence

- `pnpm test src/actions/governance/getMerklRewardsData.test.ts` → **29/29 passed**, including a new regression test `does not double-count breakdowns repeated on every page` and an updated `stops pagination at MAX_BREAKDOWN_PAGES`.
- `pnpm typecheck` → clean. `biome check` on changed files → clean (also enforced by the pre-commit hook).
- Full suite: the only failures are pre-existing **live-network** integration tests (`getMorphoVault`, `getUserVotingPowers`, `getBeamTokenLimits`, …) that fail with `Network Error`/`HTTP request failed` in the sandbox; I verified `getBeamTokenLimits` fails identically with my change stashed, so they are environmental and unrelated to this diff.

Fixes MOO-587

Linear: https://linear.app/moonwell/issue/MOO-587/check-morpho-campaign-for-safety-module

Prompted by: james-saint

### DIFF
--- a/src/actions/governance/common.ts
+++ b/src/actions/governance/common.ts
@@ -70,9 +70,23 @@ type CampaignIdsCache = {
   fetchedAt: number;
 };
 
+type MerklBreakdown =
+  MerklRewardsResponse["rewards"][number]["breakdowns"][number];
+
 const MOONWELL_MERKL_CREATOR = "0x8b621804a7637b781e2BbD58e256a591F2dF7d51";
 const CAMPAIGN_IDS_CACHE_TTL_MS = 4 * 60 * 60 * 1000; // 4 hours
 const MAX_BREAKDOWN_PAGES = 10;
+
+/**
+ * Stable identity for a Merkl reward breakdown. The Merkl `/rewards` endpoint
+ * clamps an out-of-range `breakdownPage` back to the first page instead of
+ * returning an empty page, so appending each page's breakdowns blindly
+ * re-counts the same rewards and inflates the total by the page count (up to
+ * 10x). Keying on the full breakdown content lets us skip exact repeats while
+ * still keeping genuinely distinct distributions that share a campaignId.
+ */
+const breakdownKey = (b: MerklBreakdown): string =>
+  `${b.campaignId}|${b.reason}|${b.amount}|${b.claimed}|${b.pending}|${b.subCampaignId ?? ""}`;
 
 let campaignIdsCache: CampaignIdsCache | null = null;
 
@@ -150,6 +164,14 @@ export async function getMerklRewardsData(
       return [];
     }
 
+    // Track breakdowns already collected (seeded from page 0) so repeated
+    // pages don't double-count. See breakdownKey for why Merkl repeats them.
+    const seenBreakdowns = new Set(
+      data.flatMap((reward) =>
+        reward.rewards.flatMap((r) => r.breakdowns.map(breakdownKey)),
+      ),
+    );
+
     // Paginate through remaining breakdown pages to collect all breakdowns
     for (let page = 1; page < MAX_BREAKDOWN_PAGES; page++) {
       const pageResponse = await fetch(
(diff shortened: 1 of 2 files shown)


src/actions/governance/common.ts
```diff
@@ -150,6 +164,14 @@ export async function getMerklRewardsData(
       return [];
     }
 
+    // Track breakdowns already collected (seeded from page 0) so repeated
+    // pages don't double-count. See breakdownKey for why Merkl repeats them.
+    const seenBreakdowns = new Set(
+      data.flatMap((reward) =>
+        reward.rewards.flatMap((r) => r.breakdowns.map(breakdownKey)),
+      ),
+    );
```
james-saint, line 173: Low-severity / defensive note: `seenBreakdowns` is seeded and shared **globally** across every reward entry (all tokens on the chain), but the actual append below is scoped per `(chain.id, token.address)`. So if two genuinely-distinct reward entries ever produced a breakdown with byte-identical content (`campaignId|reason|amount|claimed|pending|subCampaignId`), the second one arriving on a later page would be silently dropped — re-introducing an under-count instead of the over-count this PR fixes.

In practice this is essentially impossible (wei-precision `amount` strings colliding across two different tokens), so not a blocker. If you want to be fully robust, you could scope the set per reward entry (e.g. keyed by `token.address`) so dedup exactly mirrors the merge boundary. Fine to leave as-is.

james-saint (commented):
Overall this is a clean, well-targeted fix. The root-cause analysis is correct: Merkl clamps an out-of-range `breakdownPage` back to page 0, so the old blind-append pagination re-counted the same breakdowns up to `MAX_BREAKDOWN_PAGES` times, producing the reported ~10x inflation. Content-keyed dedup plus an early break when a page contributes no new breakdowns is the right shape, page-0 data is left untouched (so no legitimate duplicate breakdown is lost), and the new/updated tests exercise both the repeated-page regression and the hard page cap. Verified locally: `pnpm test src/actions/governance/getMerklRewardsData.test.ts` → 29/29 passed.

Findings posted inline:
- (low / defensive) `seenBreakdowns` is a single global set across all token reward entries, while the merge is scoped per `(chain, token)`; a byte-identical breakdown across two distinct entries could be dropped. Practically impossible with wei-precision amounts — noted as optional hardening, not a blocker.

No correctness blockers found.

Reviewed-at: 9d8a1e1fdce07086110163842107a146b3858d15

lyoungblood (approved):

PR #321, opened 2026-07-18
MOO-611: Read Moonriver proposal eta from the correct governor tuple index

james-saint:
## Summary

On a Moonriver proposal the timeline flipped to **"Timelock Ready to Execute"** the moment voting closed and the Execute button appeared before the timelock had elapsed.

Root cause: `getProposalsOnChainData` in `common.ts` always read the timelock `eta` from `proposalData[4]`. That index is correct for the **multichain** governor (`crossChainVoteCollectionEndTimestamp`), but the **legacy single-chain** (Moonriver) governor's `proposals()` returns a differently-shaped tuple — `(id, proposer, eta, startTimestamp, endTimestamp, …)` — where `eta` is at index **2** and index 4 is `endTimestamp` (the voting-end time, already in the past once voting closes). So the SDK reported `eta = voting-end`, making `now >= eta` true immediately.

Fix: select the tuple index by governor type — `proposalData[isMultichain ? 4 : 2]`. Verified against `governorAbi.ts` (legacy) and `multichainGovernorAbi.ts` (multichain) output orderings.

The consuming button change lives in the frontend PR on the same branch (`moonwell-frontend-v2-react`).

## Test evidence

- Added `common.test.ts` cases for the legacy governor: eta is read from index 2 (not the index-4 `endTimestamp`), and a not-yet-queued legacy proposal keeps `eta = 0` (the `votingEndTime + 1d` synthesis stays multichain-only).
- `pnpm exec vitest run src/actions/governance/proposals/common.test.ts` → **54 passed**.
- Full governance suite green except `getUserVotingPowers.test.ts`, which fails only on live-RPC `HTTP request failed` (no network in CI sandbox) and is unrelated to this change.
- `tsc --noEmit` → clean. `biome check` → clean.

Fixes MOO-611

Linear: https://linear.app/moonwell/issue/MOO-611/fix-moonriver-proposal-queue-and-execute-timeline-states

Prompted by: Luke (via Linear MOO-611)

### DIFF
--- a/.changeset/moo-611-moonriver-eta-tuple-index.md
+++ b/.changeset/moo-611-moonriver-eta-tuple-index.md
@@ -1,0 +1,5 @@
+---
+"@moonwell-fi/moonwell-sdk": patch
+---
+
+Read the Moonriver proposal `eta` from the correct `proposals()` tuple index (MOO-611). The legacy single-chain governor returns `(id, proposer, eta, startTimestamp, endTimestamp, ...)` so `eta` is at index 2, whereas the multichain governor returns `(proposer, voteSnapshotTimestamp, votingStartTime, votingEndTime, crossChainVoteCollectionEndTimestamp, ...)` where index 4 is the execution-available timestamp. The SDK was unconditionally reading index 4, so for legacy Moonriver proposals it picked up `endTimestamp` (the voting-end time, already in the past once voting closes). That made the frontend timeline flip straight to "Timelock Ready to Execute" and surface the Execute button before the timelock had actually elapsed. `eta` is now read from index 2 for legacy governors and index 4 for the multichain governor.
--- a/src/actions/governance/proposals/common.test.ts
+++ b/src/actions/governance/proposals/common.test.ts
@@ -812,3 +812,73 @@ describe("getProposalsOnChainData local read failure", () => {
     );
   });
 });
+
+// MOO-611: the legacy single-chain governor puts eta at tuple index 2, unlike
+// the multichain governor (index 4). See common.ts for the full bug context.
+describe("getProposalsOnChainData legacy single-chain governor eta (MOO-611)", () => {
+  const buildLegacyProposalsTuple = (eta: bigint) =>
+    [
+      0n, // id
+      "0x0000000000000000000000000000000000000001", // proposer
+      eta, // eta — legacy governor puts it at index 2
+      1_400_000_000n, // startTimestamp
+      1_500_000_000n, // endTimestamp — index 4, must NOT be read as eta
+      0n, // startBlock
+      0n, // forVotes
+      0n, // againstVotes
+      0n,
+      0n,
+      false,
+      false,
+    ] as const;
+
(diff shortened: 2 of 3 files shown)


james-saint (commented):
**Verdict:** LGTM. The fix correctly reads the timelock `eta` from the governor-specific tuple index (`proposalData[isMultichain ? 4 : 2]`), resolving MOO-611 where a legacy Moonriver proposal picked up `endTimestamp` (index 4) as `eta` and flipped straight to "Timelock Ready to Execute". I verified both ABIs directly: `governorAbi.ts` `proposals()` orders `(id, proposer, eta@2, startTimestamp, endTimestamp@4, ...)` and `multichainGovernorAbi.ts` orders `(proposer, voteSnapshotTimestamp, votingStartTime, votingEndTime, crossChainVoteCollectionEndTimestamp@4, ...)`, so the per-governor indices are right. The change is localized (the only index-based `proposalData` read in `src/`), the multichain synthetic-eta fallback stays multichain-only, and the added `common.test.ts` cases exercise both the queued (eta from index 2) and not-yet-queued (eta stays 0) legacy paths. Ran the focused suite: 54 passed.

- No findings — the diff looks good.

Reviewed-at: 73f3865b0bf9032eeb510bbe51b4bb5f94174009

james-saint (commented):
Re-review at 21dc2b0. The only change since my prior review (73f3865) is the new changeset `.changeset/moo-611-moonriver-eta-tuple-index.md`, which is well-formed and accurate: correct package name (`@moonwell-fi/moonwell-sdk`), `patch` bump, and a description that matches the code. I re-verified the core fix against both ABIs — legacy `governorAbi` `proposals()` returns `(id, proposer, eta, startTimestamp, endTimestamp, ...)` (eta at index 2) and `multichainGovernorAbi` `proposals()` has `crossChainVoteCollectionEndTimestamp` at index 4 — so `proposalData[isMultichain ? 4 : 2]` is correct. LGTM.

- No findings — the diff looks good.

Reviewed-at: 21dc2b0d4e64ae43ef61e9a6a6682a0bebe73b3e

apokusin (approved):
Verified the root cause against both ABIs in this repo: `governorAbi.ts` returns `(id, proposer, eta, startTimestamp, endTimestamp, …)` — `eta` at index 2, `endTimestamp` at index 4 — while `multichainGovernorAbi.ts` has `crossChainVoteCollectionEndTimestamp` at index 4. The index selector uses the same `isMultichain` flag that picks which contract was read, so shape and index can't diverge, and `proposalData` is indexed nowhere else in `src/`, so the fix is complete.

Ran the branch locally: `common.test.ts` → 54 passed, `tsc --noEmit` clean. The new tests are well-designed — the legacy tuple values differ between index 2 and 4 so the test would fail on a regression, and the second case pins the `votingEndTime + 86400` synthetic fallback as multichain-only. Patch changeset is the right bump.

Suggest cutting a release soon after merge so the frontend can drop its interim `getLegacyMoonriverEta` workaround (moonwell-frontend-v2-react#1158).

PR #324, opened 2026-08-03
feat!: remove the sunset Moonbeam/Moonriver chains, keep governance history (MOO-551)

bprofiro:
Two changes, both consequences of the Moonbeam/Moonriver sunset. **Contains a breaking change** — released as a `minor` bump per the MOO-525 precedent for breaking changes on the 0.x line, rather than jumping to 1.0.0.

---

## 1. `fix:` skip malformed Lunar market records instead of failing the chain

`fetchMarketsFromLunar` transformed every indexer record inline, so one record with a missing numeric field produced `NaN`, and `BigInt()` threw a `RangeError` that rejected **the whole chain's market list**. This hit production on the halted Moonbeam deployment once the indexer stopped backfilling it — but it is not Moonbeam-specific: any chain would behave the same way during an indexer data incident, including Base.

Each record's transform is now guarded. A malformed record is skipped with a warning and the chain's remaining markets still resolve; when *every* matched record is malformed the function throws one aggregate error, preserving the existing `getMarketsData` contract (a single meaningful `onError` plus the on-chain fallback) instead of silently returning an empty list.

Every numeric field is validated on the way in, not only the ones that reach `BigInt()`. `totalSupplyUsd`, `priceUsd`, the caps, `collateralFactor` and the APYs went through a plain `Number()`, so a dropped field landed in the returned `Market` as `NaN` — quieter than the throw-class case and worse, since one `NaN` propagates into every aggregate built from it (a chain's summed TVL) with no error raised anywhere. `toLunarNumber` routes both field classes into the same record-level guard and names the offending field, which `BigInt()`'s bare "NaN cannot be converted" never did. This half of the class was a review finding, not part of the original incident.

Reported downstream as Sentry `MOONWELL-FRONTEND-12J` / `-180` (~14.6k events / ~800 users over Aug 1–3).

## 2. `feat!:` remove the sunset chains (MOO-551)

Both chains are halted and their eRPCs are being decommissioned, so nothing on-chain can be read for them. Environments, markets, tokens, staking config and the MFAM governance token are gone — 10 definition files deleted.

### Governance history survives — this is what shaped the change

The ticket carries a ⚠️ warning that a naive removal breaks everything, and it was right for a reason worth spelling out: `getProposals` **filtered its environment list down to Moonbeam/Moonriver and returned `[]` otherwise**, and `getProposal` routed even *current* Ethereum multigov lookups **through the Moonbeam env** as the indexer source. Deleting the environments under that design would have emptied the entire proposal surface, not just the historical part.

All three actions — `getProposals`, `getProposal` and `getUserVoteReceipt` — now resolve an indexer source independently — preferring the Ethereum multigov hub, since every environment carries the same `governanceIndexerUrl` — and fan out over all three governance chains (1, 1284, 1285). On `getProposal` and `getUserVoteReceipt`, `args.chainId` now names only the chain the proposal lives on, not the environment to read through.

`getUserVoteReceipt` was a review catch: it kept resolving its environment from `args.chainId`, so post-sunset an archive lookup matched no environment and returned `[]` — a third shape alongside the didn't-vote stub and `GovernorNotFoundError`. The frontend's `useUserVoteReceipt` passes the proposal's chainId straight through, so on the bump every vote on a historical MIP would have read as "didn't vote". Its didn't-vote stub now reports the requested `chainId` rather than the hub's.

Sunset-chain proposals keep everything the indexer holds (votes, timestamps, state changes, derived state) and lose only what needed an RPC: **quorum reports 0, `eta` is unset**, and `multichain` appears only when targets identify a bridged proposal. `getProposalsOnChainData` already degraded exactly this way for unresolvable environments, so no new branch was needed.

### Two knock-on fixes on surviving chains

These are the parts I'd most like a second pair of eyes on — neither is in the ticket:

- **Hub quorum.** The local quorum read assumed a legacy `governor`. The Ethereum hub has only a `multichainGovernor`, so once Ethereum became the governance environment a hub-local proposal would have reported **quorum 0**. It now reads whichever contract the environment actually wires.
- **Native-token reward pricing.** Base and Optimism resolved their `homeEnvironment` to **Moonbeam** through `custom.governance.chainIds`, and priced native-token (`zeroAddress`) reward incentives from *Moonbeam's* views contract. With Moonbeam gone each chain resolves to itself via the existing `|| environment` fallback. The invariant test was inverted to assert no environment claims a foreign chain.

### Breaking removals

| Removed | Effect |
|---|---|
| `moonbeam` / `moonriver` chain exports, `MoonbeamEnvironment` / `MoonriverEnvironment` | compile error for direct consumers |
| `publicEnvironments.moonbeam` / `.moonriver` | `createEnvironment({ chain: moonbeam })` now throws |
| `networks.moonbeam` / `networks.moonriver` | `SupportedChains` / `SupportedChainsIds` narrow |
| `GovernanceTokensConfig.MFAM` | `GovernanceToken` narrows `"WELL" \| "MFAM"` → `"WELL"`; `WELL.chainIds` drops 1284 |

### Other behavior changes

- `getGovernanceTokenInfo` reads WELL total supply from the Ethereum hub instead of Moonbeam — **the reported figure changes** to the hub's xWELL supply rather than Moonbeam's historical total.
- `getDelegates` and `getUserVotingPowers` no longer include Moonbeam, so voting-power totals exclude any WELL that remained there.

---

## Verification

`pnpm typecheck`, `pnpm lint`, `pnpm build` and `pnpm gate:run` clean; **564 tests passing**.

Test updates reflect real contract changes rather than silenced assertions: the proposals suites now assert the three-chain fan-out and that a **1284 lookup still resolves with no Moonbeam environment present** (the guard that keeps the archive reachable), and the Ethereum invariants moved to two satellites and the inverted home-env rule. `getUserVoteReceipt.test.ts` now models the post-sunset client shape — its mock previously registered a chainId-1284 environment, which is exactly why the suite couldn't see the gap above.

Also updated: the published `src/README.md` network list, the copy-pasteable `moonClient` snippet (`moonbeam` is no longer a valid key), and the documented `governanceToken` union in two mdx pages — mirroring what MOO-525 did for `WELL_TESTNET`.

## Follow-ups, not in this PR

- **Frontend must not bump past this release yet.** `moonwell-frontend-v2-react#1190` registers both chains on a `moonwellGovernanceClient`; that client becomes *unnecessary* with this change (governance no longer needs those environments registered) and should be deleted when the frontend bumps. The frontend is pinned to `0.20.5`, so nothing breaks until then.
- Several test files still name mock objects `moonbeam`/`moonriver` (e.g. the `no lunarIndexerUrl` suites, `readCrossChainQuorums`). They build plain objects rather than importing removed code and still exercise live branches, so I left them — the names are just stale. `getUserVoteReceipt.test.ts` was the one exception where the name covered a real gap, and it is fixed here.
- `getLegacyArtemisMaxId` and the Artemis-cutoff cache are now effectively inert (no surviving environment wires a legacy `governor`, so it always returns 0). Left in place because it is written generically against `contracts.governor` and removing it would churn the classification tests; worth a separate cleanup.
- Governance docs prose still describes Moonbeam-era routing in places beyond the type unions I corrected.


bprofiro:
### Re: the two CI failures on the first push — both fixed at the source, baseline untouched

**Quality gate** flagged two metrics:

- `coverage.branches` 85.8% → 84.5%. Genuine gap: the new local-quorum selection had no tests. Added three cases (hub env with only a `multichainGovernor`, an env wiring the legacy `governor` — which must win, and an env wiring neither — quorum stays 0). Now 85.3%, within tolerance. This is the knock-on fix from the removal, so it wanted coverage regardless.
- `duplication.percentage` 8.7% → 9.0%, while `duplication.clones` held at **93** and `duplicatedLines` actually *fell* 1903 → 1901. Purely an arithmetic artifact — deleting the sunset definitions shrank the denominator by ~750 lines that were never part of any clone.

  I did not re-baseline that. Instead I removed duplication jscpd had been flagging all along: `getProposal` and `getProposals` each carried their own copy of the `ApiProposal → Proposal` mapping — 69 duplicated lines — and *each carried a comment warning the other not to drift from it*. Extracted `mapApiProposalToProposal` into `common.ts`, so there is one implementation and that drift is now impossible. Duplication is **1834 lines / 91 clones**, below the committed baseline.

**Cloudflare Pages** failed because vocs type-checks every mdx code sample, and ten governance pages imported `moonbeam` from the environments barrel. They now demonstrate the Ethereum multigov hub, and the shared `moonClient` snippet registers `ethereum` so `network: "ethereum"` type-checks. Verified with a local `vocs build`.

`quality-gate/baseline.json` is unchanged.

bprofiro:
## Summary

Two independent changes. The first guards each Lunar record's transform so one malformed record no longer rejects a whole chain's market list — a genuinely good fix with the right escalation (aggregate throw when *every* record is bad, preserving `getMarketsData`'s on-chain fallback). The second removes the sunset Moonbeam/Moonriver chains and re-homes the governance surface on the Ethereum multigov hub.

The public API impact is real and the Changeset captures it accurately: chain exports, environment types, `publicEnvironments` entries, `networks.*` keys and `GovernanceTokensConfig.MFAM` all disappear, narrowing `SupportedChains`, `SupportedChainsIds` and `GovernanceToken`. The `minor`-not-`major` bump follows the MOO-525 precedent. The two knock-on fixes the description flagged for scrutiny — hub quorum source and native-token reward pricing — both look correct to me, and the inverted `homeEnvironment` invariant test is a stronger assertion than the one it replaced.

The findings below cluster on one theme the PR under-covers: **what the archive chains now do at runtime once no environment exists for them.**

---

## Issues

### 🔴 Critical

#### 1. Removed network keys silently build a Polygon environment — [`src/client/createMoonwellClient.ts:89`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/client/createMoonwellClient.ts#L89)

The network-key ternary ends in a bare `: polygon` fallthrough. With the `moonbeam`/`moonriver` branches deleted, those keys no longer match anything and fall through to Polygon:

```ts
curr === "ethereum" ? ethereum : curr === "avalanche" ? avalanche : curr === "arbitrum" ? arbitrum : polygon
```

A JS consumer calling `createMoonwellClient({ networks: { moonbeam: { rpcUrls: ["https://rpc.moonwell.fi/main/evm/1284"] } } })` gets a **Polygon environment keyed `moonbeam`**, wired with Polygon contract addresses over a Moonbeam RPC. Every read against it returns garbage or reverts, with no error naming the cause.

This matters concretely: the follow-up section notes `moonwell-frontend-v2-react#1190` registers both chains on a `moonwellGovernanceClient`. TypeScript narrows `NetworksConfig` to `SupportedChains`, so a typed consumer is protected — but JS consumers, `as any` call sites, and config built from runtime data are not. The changeset promises `createEnvironment({ chain: moonbeam })` "now throws"; that guarantee doesn't hold on this path.

```text
In src/client/createMoonwellClient.ts, the network-key ternary in createMoonwellClient ends with a bare `: polygon` fallthrough, so any unrecognised network key (including the removed "moonbeam"/"moonriver") silently builds a Polygon environment. Replace the trailing fallthrough with an explicit `curr === "polygon" ? polygon : ...` branch and throw an Error naming the unsupported network key instead, matching the `throw new Error("Unsupported chainId")` behaviour of createEnvironment in src/environments/index.ts. Add a test asserting createMoonwellClient throws for a "moonbeam" network key rather than returning a Polygon environment.
```

#### 2. Archived proposals that were Defeated now surface as `Pending` — [`src/actions/governance/proposals/common.ts:463`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/proposals/common.ts#L463)

With no 1284/1285 environments left, `resolveHomeEnv` returns `undefined` for every archive proposal, so the `!homeEnv` branch is now their **permanent** route rather than a transient degradation. That branch derives state via `deriveProposalStateFromApi`, which can only return `Executed → Canceled → Queued → Active → Pending`.

A historical Moonbeam proposal that was **Defeated** — voting ended, no `QUEUED`/`CANCELED`/`EXECUTED` state change — matches nothing, falls past the voting-window check, and returns `Pending`. Every defeated proposal in the archive lists as perpetually Pending; bridged ones additionally carry `multichain.votesCollected: false`. Pre-sunset these read real `Defeated` state from the Moonbeam governor.

The PR description says sunset-chain proposals "keep everything the indexer holds (votes, timestamps, state changes, derived state)" and lose only quorum/eta. Defeated-rendered-as-Pending is a third loss, and it's the one most visible in the UI.

```text
In src/actions/governance/proposals/common.ts, deriveProposalStateFromApi cannot express ProposalState.Defeated, so archived Moonbeam/Moonriver proposals that were defeated now permanently surface as Pending (the !homeEnv branch is their only route now that no environment exists for chains 1284/1285). Extend deriveProposalStateFromApi so that when the voting window has closed and there is no QUEUED/CANCELED/EXECUTED state change, it decides between Succeeded and Defeated using the API proposal's forVotes/againstVotes tallies rather than falling through to Pending. Keep the existing precedence for the terminal states. Add tests in src/actions/governance/proposals/common.test.ts covering a past-voting-window proposal with more against than for votes (expect Defeated) and the reverse (expect Succeeded).
```

#### 3. `getProposals({ chainId: 1284 })` throws a `TypeError` — [`src/actions/governance/proposals/getProposals.ts:38`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/proposals/getProposals.ts#L38)

`chainId` is a public parameter of `getProposals` (via `OptionalNetworkParameterType`). When it matches no registered environment, `getEnvironmentsFromArgs` takes its `Number.isInteger` branch and returns `[undefined]` — a one-element array holding `undefined`, not an empty array:

```ts
return [Object.values(client.environments).find((env) => env.chainId === chainId)] as Environment[];
```

`resolveGovernanceEnvironment` then runs `environments.find((e) => e.chainId === mainnet.id)` and dereferences `undefined` → `Cannot read properties of undefined (reading 'chainId')`.

The crash isn't new for arbitrary chain ids, but it *is* newly guaranteed for **1284 and 1285** — which were registrable before this PR and are exactly the ids `getProposal`'s docblock teaches callers to pass ("the chain the proposal lives on"). A consumer reasonably mirroring that parameter across the two actions gets a hard crash on the archive chains this PR is built to keep reachable. The `!governanceEnvironment` guard right below never gets the chance to fire.

```text
In src/actions/governance/proposals/getProposals.ts, getProposals passes the result of getEnvironmentsFromArgs straight into resolveGovernanceEnvironment. When args.chainId matches no registered environment, getEnvironmentsFromArgs returns [undefined], and resolveGovernanceEnvironment throws a TypeError dereferencing it. Filter falsy entries out of the environments array before calling resolveGovernanceEnvironment so the existing `if (!governanceEnvironment) return []` guard handles the case. Do the same defensively wherever else the result of getEnvironmentsFromArgs is dereferenced. Add a test asserting getProposals(client, { chainId: 1284 }) on a client with only an ethereum environment resolves rather than throwing.
```

---

### 🟡 Major

#### 4. Partially malformed Lunar data never reaches `onError` — [`src/actions/core/markets/common.ts:744`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/core/markets/common.ts#L744)

The skip path emits only a `console.warn`. Both the all-malformed case and the total-failure case route through `environment.onError` — but the partial case, which is the *most likely* real incident shape, reports nothing to Sentry-wired consumers.

If an indexer incident corrupts 3 of 20 Base records, three markets silently vanish from `getMarkets` output and the only trace is a browser console line. That's the MOONWELL-FRONTEND-12J class of incident becoming invisible rather than visible — the inverse of what this fix is for. The codebase's own convention (the quorum reads, `getWellPriceFromBaseOracle`) is to route swallowed failures through `onError` precisely so they stay observable.

```text
In src/actions/core/markets/common.ts, fetchMarketsFromLunar only console.warns when it skips a malformed record — environment.onError is never called for the partial-failure case, so a partial indexer incident is invisible to Sentry-wired consumers. After the record loop, when malformedCount > 0 but markets.length > 0, call environment.onError with an Error summarising how many records were skipped on which chain, using a distinct source string (e.g. "markets-malformed-records") so it is attributable separately from the existing "markets" source. Add a test in src/actions/core/markets/common.test.ts asserting onError fires once with that source when one of two records is malformed, and that the valid market is still returned.
```

#### 5. One bad incentive field discards an otherwise-valid market — [`src/actions/core/markets/common.ts:550`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/core/markets/common.ts#L550)

The `try` opens before the core numerics and closes after the incentives loop, so it covers three `BigInt()` calls on incentive fields (`borrowIncentivesPerSec` at :671 and :697, `supplyIncentivesPerSec` at :698). A record whose market numerics are all valid but whose incentive carries `supplyIncentivesPerSec: null` throws inside that loop, and the catch discards the **whole market**.

During a rewards-field-only indexer incident this removes every market from the chain rather than degrading their APRs — precisely the blast radius the PR set out to shrink. The incentives loop already has a `continue`-on-unusable-entry idiom (`if (!token) continue`, `if (!price) continue`); an inner guard would fit it naturally.

```text
In src/actions/core/markets/common.ts, the try block in fetchMarketsFromLunar wraps both the core market numerics and the incentives loop, so a malformed incentive field (BigInt on a null borrowIncentivesPerSec / supplyIncentivesPerSec) discards the entire market record. Narrow the blast radius: wrap the body of the `for (const incentive of lunarMarket.incentives)` loop in its own try/catch that skips just that incentive and warns, matching the existing `if (!token) continue` / `if (!price) continue` idiom, so the market is still returned with its remaining rewards. Keep the outer try for the core numerics. Add a test asserting a record with one malformed incentive still yields the market, with the bad reward entry omitted.
```

#### 6. Non-null assertions re-introduced in library source — [`getProposal.ts:110`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/proposals/getProposal.ts#L110), [`getProposals.ts:131`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/proposals/getProposals.ts#L131)

`CLAUDE.md` states: *"No non-null assertions (`!`) — use optional chaining or nullish coalescing."* Both call sites were rewritten by this diff and carry `onChainDataList[0]!` / `onChainDataList[index]!` in `src/` (the rule allows `!` only in test files). `getProposalsOnChainData` maps 1:1 over its input, so the assertions are *true* — but the rule is unconditional, and a guard expresses the invariant rather than asserting it away.

```text
In src/actions/governance/proposals/getProposal.ts and src/actions/governance/proposals/getProposals.ts, the calls to mapApiProposalToProposal use non-null assertions (onChainDataList[0]! and onChainDataList[index]!), which the repo CLAUDE.md forbids in src/ ("No non-null assertions (!) — use optional chaining or nullish coalescing"). Restructure both to avoid the assertion: in getProposals, iterate with a guard that skips (or filters) entries where the corresponding onChainData is undefined; in getProposal, return undefined when onChainDataList[0] is missing. Do not change the 1:1 mapping contract of getProposalsOnChainData.
```

#### 7. `getGovernanceTokenInfo` silently ignores `governanceToken` — [`src/actions/governance/getGovernanceTokenInfo.ts:24`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/getGovernanceTokenInfo.ts#L24)

The parameter is retained for API compatibility but no longer read. A JS consumer still passing `{ governanceToken: "MFAM" }` — which previously returned Moonriver's MFAM total supply — now receives the Ethereum **xWELL** supply as a normal success, with nothing distinguishing it from a correct call. Keeping the parameter shape while dropping its meaning converts a compile error (for TS consumers) into silently wrong data (for everyone else).

```text
In src/actions/governance/getGovernanceTokenInfo.ts, the governanceToken argument is retained for API compatibility but ignored, so a runtime caller passing "MFAM" silently receives WELL's supply. Add a runtime guard that throws (or returns undefined, matching the declared GetGovernanceTokenInfoReturnType) when args.governanceToken is provided and is not "WELL", with an error message explaining MFAM was removed with the Moonriver sunset. Rename the parameter back from _args to args. Add a test covering the "MFAM" input.
```

---

### 🔵 Minor

#### 8. `SUPPORTED_GOVERNOR_CHAIN_IDS` docblock now contradicts its own consumers — [`src/actions/governance/governor-api-client.ts:18`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/governor-api-client.ts#L18)

The comment says *"This is the fan-out set for `getUserVoteReceipt`"* and warns that *"reaching a 1285 proposal through a Moonbeam env would surface a degraded result"*. This PR makes `getProposals` fan out over the same set, deletes every Moonbeam env, and deliberately serves 1285 through a foreign env with `quorum: 0`. The next editor following this comment will either "fix" the new fan-out or misread why explicit `chainId: 1285` lookups degrade.

```text
In src/actions/governance/governor-api-client.ts, update the docblock on SUPPORTED_GOVERNOR_CHAIN_IDS: it claims the constant is the fan-out set for getUserVoteReceipt only, and warns against reaching a 1285 proposal through a foreign environment — both statements are now wrong. getProposals fans out over the same set, no Moonbeam environment exists, and serving 1285 through the resolved governance environment (with quorum 0 and no eta) is the accepted post-sunset design. Rewrite the comment to name both consumers and state the degradation as intended behaviour, keeping the distinction from MULTIGOV_PROPOSAL_FALLBACK_CHAIN_IDS.
```

#### 9. `buildProposals` docblock names deleted callers — [`src/actions/governance/proposals/getProposals.ts:109`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/governance/proposals/getProposals.ts#L109)

*"Used by both the Moonbeam/Ethereum and Moonriver paths so the mapping stays in one place"* — this diff deleted `getMoonbeamProposals` and `getMoonriverProposals`; the function now has exactly one caller.

```text
In src/actions/governance/proposals/getProposals.ts, the docblock above buildProposals says it is "Used by both the Moonbeam/Ethereum and Moonriver paths", but this change deleted getMoonbeamProposals and getMoonriverProposals and buildProposals now has a single caller (fetchGovernorProposals). Update the comment to describe the current single-caller pipeline. Consider whether buildProposals still earns being a separate function or should be inlined into fetchGovernorProposals.
```

#### 10. Published README contradicts itself on supported networks — [`src/README.md:43`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/README.md#L43)

Line 43 says *"Support for multiple networks: Base, Optimism, and Ethereum"*; the Supported Networks section at lines 81–86 lists six, including Arbitrum, Avalanche and Polygon. The same edit updated one list but not the other, and this file ships in the published package.

```text
In src/README.md, line 43 says "Support for multiple networks: Base, Optimism, and Ethereum" while the Supported Networks section at lines 81-86 lists six chains (Base, Optimism, Ethereum, Arbitrum, Avalanche, Polygon). Update line 43 so both lists agree.
```

---

## Positive observations

- **The malformed-record escalation is exactly right.** Skipping bad records but throwing an aggregate error when *every* record fails preserves `getMarketsData`'s existing contract — one meaningful `onError` plus the on-chain fallback — instead of silently returning `[]` and making an indexer outage look like an empty protocol. The `malformedCount > 0 && markets.length === 0` condition is the correct guard, not `markets.length === 0` alone.
- **`mapApiProposalToProposal` deletes a real drift hazard.** Both copies carried comments warning the *other* not to drift — the strongest possible signal that a single implementation was overdue. Routing `isMultichain` through `onChainData` rather than re-classifying keeps the classification and the read-routing provably in sync.
- **The inverted `homeEnvironment` invariant test is a genuine upgrade.** `expect(claimed).toEqual([])` asserts a general rule ("no environment claims a foreign chain") where the old test asserted one instance, and it will catch the next mis-wiring rather than only the Moonbeam one. Same for the explicit `.not.toContain(moonbeam.id)`.
- **The hub-quorum fix is a real catch and correctly ordered.** Preferring `governor.getQuorum()` and falling back to `multichainGovernor.quorum()` keeps legacy envs on their existing path while fixing hub-local proposals reporting quorum 0 — and the three new tests cover all three branches including the neither-wired case.
- **The 1284-lookup-with-no-Moonbeam-env test is the right regression guard**, and pins precisely the invariant that makes the archive reachable.
- The PR description flagging the two out-of-ticket knock-on fixes for scrutiny is exactly the right call — both were worth the second look.

---

## Suggestions

#### The `custom.governance.chainIds` home-environment mechanism is now provably dead

Every surviving environment ships `chainIds: []`, and the new invariant test forbids any environment from claiming a foreign chain. So this lookup — at [`markets/common.ts:41`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/core/markets/common.ts#L41), [`markets/common.ts:503`](https://github.com/moonwell-fi/moonwell-sdk/blob/brenda/fix-lunar-malformed-market-records/src/actions/core/markets/common.ts#L503), and `user-rewards/common.ts:21` — can never match, and always falls through to `|| environment`:

```ts
const homeEnvironment =
  (Object.values(publicEnvironments) as Environment[]).find((e) =>
    e.custom?.governance?.chainIds?.includes(environment.chainId),
  ) || environment;
```

Relying on the `|| environment` fallback is correct and safe for this PR — but it leaves resolution machinery, an `Object.values(publicEnvironments)` scan on a hot path, and an invariant test all guarding a branch that cannot fire. Worth a follow-up to either collapse the call sites to `environment` and retire the field, or document it as a deliberate extension point. (Same shape as the `getLegacyArtemisMaxId` inertness already noted in the follow-ups.)

```text
With Moonbeam removed, every surviving environment has custom.governance.chainIds: [], and the ethereum environment.test.ts invariant now forbids any environment from claiming a foreign chain — so the homeEnvironment lookups in src/actions/core/markets/common.ts (around lines 41 and 503) and src/actions/core/user-rewards/common.ts always fall through to `|| environment`. Investigate whether custom.governance.chainIds still has any live consumer. If not, collapse those call sites to use `environment` directly, remove the now-dead Object.values(publicEnvironments) scans, and decide whether the chainIds field and its invariant test should be retired or kept with a comment documenting it as an intentional extension point. Do not change behaviour — the fallback result is already identical.
```

#### Add archive-path coverage keyed on "no environment exists"

Findings 2 and 3 share a root: the tests exercise the archive through mocked `getProposalsOnChainData`, so nothing covers the real `!homeEnv` branch that every 1284/1285 proposal now takes. A test that runs `getProposalsOnChainData` unmocked against a client holding only an Ethereum environment, with a 1284 proposal in various terminal states, would have caught both.

```text
Add a test suite in src/actions/governance/proposals/common.test.ts that exercises getProposalsOnChainData unmocked for archive-chain proposals (chainId 1284 and 1285) against a governance environment for chain 1 only — the real post-sunset configuration, where resolveHomeEnv returns undefined and the !homeEnv branch is the permanent route. Cover proposals in each terminal state (executed, canceled, defeated, succeeded-but-unexecuted) and assert the derived state, quorum and eta match what the PR documents. Remember to add any new test file to the include array in test/vitest.config.ts.
```

bprofiro:
## Review addressed — all 10 findings + both suggestions (`2d70834`)

The review was right that the PR under-covered **what the archive chains actually do at runtime once no environment exists for them**. Two of the three criticals were that theme, and the fix for both is now pinned by a suite that exercises the real post-sunset configuration instead of mocking past it.

`quality-gate/baseline.json` is unchanged. **556 tests passing**, up from 527.

---

### 🔴 Critical

**#1 — Removed network keys silently build a Polygon environment.** Every supported key now has its own branch in `createMoonwellClient`; anything else throws `Unsupported network "<key>"`, matching `createEnvironment`'s `Unsupported chainId`. The changeset's promise that `createEnvironment({ chain: moonbeam })` throws now holds on this path too. Tests cover both removed keys, an arbitrary unknown key, and that all six real keys still build their own environment (the regression the explicit branches could introduce).

**#2 — Archived Defeated proposals surface as `Pending`.** `deriveProposalStateFromApi` now decides `Succeeded` / `Defeated` from the tallies once the voting window has closed with no `QUEUED`/`CANCELED`/`EXECUTED` change. Existing precedence is unchanged, and a terminal state change still wins over the tallies — there's a test for that specifically.

One caveat worth stating rather than hiding: **quorum needs an RPC and reports 0 for the archive**, so a for-majority that missed quorum reads as `Succeeded`, not `Defeated`. There's no signal available on this path to distinguish them. That's the same degradation `quorum` already carries, and it's strictly closer to the truth than `Pending` — so it's documented on the function and called out in the changeset instead of being papered over.

**#3 — `getProposals({ chainId: 1284 })` throws a `TypeError`.** Fixed at the source in `getEnvironmentsFromArgs` rather than at the one call site: it returned `[undefined]`, and every one of its ~18 callers dereferences what it gets back, so the same crash was reachable from `getMarkets`, `getUserPositions` and the rest. It now returns `[]`, which lets each action's existing "nothing to do" guard fire — including the `if (!governanceEnvironment) return []` right below the crash site. Removing the `[undefined]` shape also deleted two `as Environment[]` casts.

### 🟡 Major

**#4 — Partial malformed Lunar data never reaches `onError`.** Skipped records now report through `environment.onError` under `markets-malformed-records`, separately attributable from the total failure that reports under `markets`. The partial case is the one that returns *successfully*, which is exactly why it needed its own report.

**#5 — One bad incentive field discards an otherwise-valid market.** The incentives loop body has its own try/catch now, so a malformed reward costs its own entry and the market is still returned with its remaining rewards. Skipped incentives report under `markets-malformed-incentives`.

**#6 — Non-null assertions.** Both replaced with guards: `getProposals` maps through a `flatMap` that drops unmatched entries, `getProposal` returns `undefined`. `getProposalsOnChainData`'s 1:1 contract is unchanged — each guard has a test that feeds a short list, so the branch is exercised rather than merely asserted.

**#7 — `getGovernanceTokenInfo` silently ignores `governanceToken`.** Throws for anything other than `"WELL"`, naming MFAM and the Moonriver sunset. Parameter renamed back from `_args`. The test file wasn't in the vitest `include` array, so it's been added — the pre-existing WELL test now runs too.

### 🔵 Minor

**#8, #9, #10** — `SUPPORTED_GOVERNOR_CHAIN_IDS` now names both consumers and states the 1284/1285 degradation as intended behaviour while keeping the `MULTIGOV_PROPOSAL_FALLBACK_CHAIN_IDS` distinction; `buildProposals`'s docblock describes its single caller; `src/README.md` line 43 lists all six networks.

---

### Suggestions

**Archive-path coverage** — added, and it's what pins #2 and #3. `getProposalsOnChainData` runs unmocked against a governance environment for chain 1 only, with 1284/1285 proposals in each terminal state: executed, canceled, defeated, succeeded-but-unexecuted, queued, and mid-window. It also asserts the things the PR description promised — `proposalData: null`, `quorum: 0n`, `eta: 0` — plus that `isMultichain` stays false, since with no env there's no multichain governor for the unknown-cutoff bias to route to.

**`custom.governance.chainIds` is provably dead** — confirmed: every surviving environment ships `[]`, and the six lookups always fall through to `|| environment`. I documented it as a deliberate extension point on the type rather than collapsing the call sites, because retiring the field changes `Environment["custom"]` — another breaking change in a PR that already carries several, touching morpho paths this PR otherwise doesn't. Better as its own cleanup, alongside the `getLegacyArtemisMaxId` inertness already in the follow-ups.

---

### Quality gate

Branches came in at **85.4%** against the 85.8% baseline, inside tolerance. It failed at 85.1% on the first pass — the guards from #6 and the optional-chain `onError` calls from #4/#5 add branches that success-path tests don't reach — so I covered them (short-list guards, an environment wiring no `onError`, an omitted `governanceToken`) rather than move the floor. Duplication holds at 91 clones / 1835 lines, still below baseline.

Both changesets updated: the runtime guards from #1/#3/#7 and the archive state change from #2 are all consumer-visible.

apokusin (changes requested):
Reviewed the full diff plus surrounding source, ran the six touched suites locally (142/142 passing; CI green — my only local failures were live-RPC integration flakes in untouched Morpho suites). Both knock-on fixes you asked for eyes on check out:

- **Hub quorum**: the source selection reads whichever contract the env actually wires — legacy `governor` preferred, hub `multichainGovernor` otherwise — and the three `local quorum source selection` tests pin each branch. Verified `readCrossChainQuorums` and `resolveHomeEnv` still resolve satellites through `publicEnvironments`, so hub quorum works even on a client that never registered Ethereum.
- **Native-token reward pricing**: the fix is correctly the *deletion* of Moonbeam's foreign `custom.governance.chainIds` claim — Base/Optimism/Ethereum already shipped `[]`, so each chain falls back to itself via the existing `|| environment` at both call sites (`user-rewards/common.ts`, `markets/common.ts:41,503`). The inverted invariant ("no environment claims a foreign chain") is the right shape to keep it that way.

I also confirmed the doc claim that every surviving environment defaults the same `governanceIndexerUrl`, so `resolveGovernanceEnvironment`'s `?? environments[0]` fallback is safe on any client shape.

Requesting changes for one gap that gates the frontend bump:

## 1. `getUserVoteReceipt` still routes the environment through the proposal's chainId — archive receipts silently vanish

The exact design flaw this PR fixes in `getProposal` ("`args.chainId` names the chain the proposal lives on, not the environment to read through") survives in `getUserVoteReceipt`:

```ts
// src/actions/governance/getUserVoteReceipt.ts:53
const environment = getEnvironmentFromArgs(client, args);
if (!environment) {
  return [];
}
```

Its own param doc (lines 24–31) teaches callers that `chainId: 1284` / `1285` are valid values. Post-sunset no registrable environment carries those chainIds, so an explicit archive lookup returns `[]` — not the `voted: false` stub, not `GovernorNotFoundError`, but a third shape callers don't handle.

This is reachable, not theoretical: the frontend's `useUserVoteReceipt` hook passes the proposal's chainId straight through (`moonwell-frontend-v2-react/src/hooks/useUserVoteReceipt.tsx:30`), and archive proposals served by this PR's `getProposal` carry `chainId: 1284/1285`. The moment the frontend bumps to this release, every vote receipt on a historical MIP reads as "didn't vote."

Fix is the same move you already made in `getProposal`: resolve the indexer source via `resolveGovernanceEnvironment(Object.values(client.environments))` and keep `args.chainId` purely as the proposal-chain selector for `tryChains`. Happy with it in this PR or a fast-follow, but it belongs on the "frontend must not bump past this release yet" list either way.

## 2. One "stale mock name" is actively masking finding 1

The PR body's "the names are just stale" holds for the suites you listed — except `getUserVoteReceipt.test.ts`, where the mock client *registers an environment with chainId 1284* (lines 23–30). That's a production shape that can no longer exist, and it's precisely why the suite can't see the gap above. When fixing #1, that mock client should model the post-sunset shape (Ethereum env only).

## 3. Optional follow-up, not blocking

The markets guard catches throw-class malformation (`BigInt(NaN)`) — the right scope for this incident. A record whose bad field feeds a plain `Number()` (`totalSupplyUsd`, `priceUsd`, APYs) still passes through as silent `NaN`. Pre-existing, but if the indexer can drop one field class it can drop the other; up-front record validation would retire the whole class.

---

Everything else is genuinely strong: the malformed-record guard has exactly the right blast radius (record-level + incentive-level guards, aggregate throw preserving the on-chain fallback contract, distinct `onError` sources so partial degradation is observable), the tally-derivation branch is correct (bigint compare, tie → Defeated matching governor strictly-greater semantics, terminal events still win, the missed-quorum→Succeeded degradation documented rather than hidden), `createMoonwellClient` failing loudly on unknown keys retires a real footgun, and the changesets/docs/PR-body honesty about the WELL-supply and delegate-total changes is exemplary.

bprofiro:
Thanks — all three addressed in 8ff93f3, including the optional one.

## 1. `getUserVoteReceipt` — fixed

You're right that it's the same flaw, and the reachability argument through `useUserVoteReceipt` is what made me fix it here rather than fast-follow: the PR that makes archive proposals readable shouldn't be the one that makes their receipts unreadable.

Resolved via `resolveGovernanceEnvironment(Object.values(client.environments))`, with `args.chainId` left as a pure `tryChains` selector — same shape as `getProposal`.

One thing that fell out of the fix and wasn't in your finding: the didn't-vote stub was built from `environment.chainId`, i.e. the args-resolved environment. Swapping in the governance environment unqualified would have made every archive stub claim `chainId: 1`, trading a wrong `voted` for a wrong chain. It now reports `chainId ?? governanceEnvironment.chainId`, so an explicit 1284 lookup gets a 1284 stub and a bare lookup gets the source it actually read through. Pinned by its own test.

## 2. The masking mock — fixed

Good catch, and the distinction you drew is the right one: the other suites' names are cosmetic, this one encoded an impossible client shape and that's precisely why the suite was blind. The mock client is now Ethereum-only, and the `network: "moonbeam"` args are gone throughout.

Two consequences worth flagging:

- The old `returns [] when the requested env doesn't exist in the client` test lost its premise entirely — env resolution no longer reads args, so `network: "polygon"` is now simply ignored. It became an empty-environments guard, mirroring `getProposal.test.ts`.
- Added `resolves the indexer source independently of the requested chainId` and `the didn't-vote stub reports the requested proposal chain, not the hub`. Both fail against the pre-fix code, so the gap is now covered rather than just closed.

## 3. The `Number()` half — also fixed here

I took this in-PR rather than as a follow-up. Your framing — "if the indexer can drop one field class it can drop the other" — held up, and on a closer look the silent half is the worse one: a `NaN` `totalSupplyUsd` doesn't stay in its own record, it propagates into any aggregate built from it, so one bad market turns a chain's summed TVL into `NaN` with nothing raised anywhere. The throw-class case at least failed loudly.

`toLunarNumber(value, field)` now backs all 26 numeric call sites, so both classes land in the same record-level guard. The record is skipped and reported through the existing `markets-malformed-records` source, and the error names the field — `BigInt()`'s bare "NaN cannot be converted to a BigInt" told you a record was bad but never which field. Incentive-level tiering is preserved: a `NaN` APR still costs only its own reward entry.

I chose skip-over-report deliberately. Returning the market with `NaN` in one cell and an `onError` alongside would preserve more data, but it leaves the poisoning behavior intact for anything summing the list — and the record-level skip is the blast radius this PR already argues for.

Covered by a `test.each` over `totalSupplyUsd` / `priceUsd` / `collateralFactor` / `baseSupplyApy` asserting no `NaN` survives into the returned market, a case asserting the error names the field, and one for the incentive-APR tiering.

## Verification

564 tests passing (was 527); `typecheck`, `lint`, `build` and `gate:run` clean — gate passes with coverage up (+0.27pp lines, +0.81pp functions), branches -0.22pp within tolerance. Both changesets and the PR body updated to describe the new behavior.

The "frontend must not bump past this release yet" note stands, but for the `moonwellGovernanceClient` deletion rather than for vote receipts — finding 1 was the item that would have made the bump actively lossy.

apokusin (approved):
Verified commit 8ff93f3 against my previous review — all three findings addressed correctly:

1. **`getUserVoteReceipt`** now resolves the indexer source via `resolveGovernanceEnvironment` instead of `getEnvironmentFromArgs(client, args)`, so an explicit archive lookup (`chainId: 1284/1285`) reaches the hub instead of matching no environment. You also caught the corollary I missed: the didn't-vote stub previously always reported `environment.chainId`, which would have mislabeled an archive stub as the hub's chain — now `chainId ?? governanceEnvironment.chainId` reports what was actually asked for.
2. **The mock client** in `getUserVoteReceipt.test.ts` is now the real post-sunset shape (Ethereum-only), with `resolves the indexer source independently of the requested chainId` and `the didn't-vote stub reports the requested proposal chain, not the hub` pinning exactly the regression.
3. **`toLunarNumber`** closes the silent-NaN gap on the `Number()`-typed fields — good call routing it through the same record-level try/catch rather than a second failure path, and naming the field beats `BigInt()`'s bare message for debugging the next incident.

Ran the seven touched suites locally against this commit (158/158 passing) and confirmed CI (Lint, Test, gate) is green.

Approving.

